PR #157, opened 2015-12-04
Allow passing a string as argument to the Lua startup file

timfel:
This is related to https://github.com/Wargus/wargus/issues/132, as this allows us to handle commandline arguments in lua scripts. This commit also makes the script searching a little more clever, so we could conceivably have lua scripts in the user directory that then handle arguments and run the game.

### DIFF
--- a/src/include/parameters.h
+++ b/src/include/parameters.h
@@ -46,6 +46,7 @@ public:
 	std::string applicationName;
 	std::string luaStartFilename;
 	std::string luaEditorStartFilename;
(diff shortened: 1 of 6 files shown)


Andrettin:
Nice idea :)

Maybe instead of having two functions using luaArgStr = "" could be a good idea instead?

timfel:
Done. From my side this is good to go. Any other remarks @Andrettin?

Andrettin:
No further remarks, I agree that this is good to go.

PR #160, opened 2015-12-10
For incomplete wall tilesets, keep the current directionality if the new direction is missing

timfel:
I wpuld like to use this for war1gus, where dungeon doors are destructible like walls, but they dont have all the possible directions. @Andrettin, you have dungeons in Wyrmsun, too, do you think this looks ok, or might it break something?


Andrettin:
Hey Tim, I've tested these changes with Wyrmsun and I couldn't find any issues. It looks ready to go for me :)

Mailaender:
This spits out a compiler warning that rpmlint classifies as "serious". Filed a fix at https://github.com/Andrettin/Wyrmgus/pull/50.

PR #162, opened 2015-12-14
Porting to SDL2

timfel:
This was already shortly discussed back on Launchpad, but it doesn't seem it has gotten anywhere.

(sometimes Markdown is useful, but sometimes it plain sucks)

 I've got a semi-serious patch, that already produces decent results, but it has a few issues:
- flipped sprites are invisible for no apparent reason 
- wasteland, winter and swamp tilesets work, but summer doesn't show the terrain (results in cursor artefacts all over the visible map)
- cursor wipes out text on widgets
- text input is yet to be fixed properly (right now, anything not ASCII will fail)
- while fullscreen, the mouse is clipped to 640x480 window

[stratagus_2.3.0-sdl2-x2.patch.txt](https://github.com/Wargus/stratagus/files/60363/stratagus_2.3.0-sdl2-x2.patch.txt)

(on a semi-related note: refusing to attach a patch, unless it has .txt extension ? seriously !?)

A block or two might have been taken from the Launchpad branch (IIRC (it's been awhile since I've started writing this patch) the mouse wheel change comes from there).

I've simply disabled OpenGL code for the time being.

As the name says, the patch is against 2.3.0.


timfel:
@galtgendo Thanks for those patches! If you would rather not create a pull request, that is fine by me. I have taken the liberty to import your patches on a branch and convert this issue to a pull request from that branch, where any further patches can be integrated. This way, we at least keep history of your efforts and we can browse code changes

galtgendo:
OK, so about that build failure, obviously, I didn't include FindSDL2.cmake, cause it was already in your old branch **AND** there are many versions of it floating around.

galtgendo:
@Andrettin I didn't touch frame_map generation and the picture itself is correctly generated...

Though it seems the palette is getting trashed: palettes taken from slightly after LoadGraphicPNG(this) completes in CGraphic::Load and at the end of CGraphic::Flip for Surface are equal, yet palette for SurfaceFlip, also taken at the end of CGraphic::Flip differs.

galtgendo:
...and obviously those SDL_{G,S}etRelativeMouseMode calls should actually be 'SDL_{G,S}etWindowGrab(TheWindow' calls - got that confused with another port, that actually needed RelativeMouseMode.

Andrettin:
@galtgendo I think I may have found an issue with the flip code:

```
if (!SDL_GetColorKey(Surface, &ckey)) {
		SDL_SetColorKey(SurfaceFlip, SDL_TRUE, ckey);
```

Shouldn't it be:
```
if (SDL_GetColorKey(Surface, &ckey)) {
		SDL_SetColorKey(SurfaceFlip, SDL_TRUE, ckey);
```

Instead?

galtgendo:
@Andrettin if it were that easy:
> return 0 on success, or -1 if the surface is not valid or colorkey is not enabled

galtgendo:
In the meanwhile minor fixes: in src/video/sdl.cpp, in SdlDoEvent

    if (InMainWindow && (event.window.event == SDL_WINDOWEVENT_LEAVE)) {
        InputMouseExit(callbacks, SDL_GetTicks());
    }
    InMainWindow = (event.window.event == SDL_WINDOWEVENT_ENTER);

and

    event.window.event == SDL_WINDOWEVENT_FOCUS_LOST

    event.window.event == SDL_WINDOWEVENT_FOCUS_GAINED

instead of not really correct '&' checks - that makes (among other) pause on unfocus work properly again.

galtgendo:
...no real progress this time, just correcting myself - palettes weren't getting trashed at the end of CGraphic::Flip, my debug block was counting wrong (so badly that I'm a bit surprised I didn't get a segfault for reading past bounds).

galtgendo:
Well, this time some real progress: by some "printf debugging" I found the reason for the trouble with flipped sprites - I'd call it a policy change in SDL2 paletted surface handling.

If an indexed surface is blited onto non-indexed, the index surface can at most have color key set.
AlphaMod is no-go, but also there's that little bit that SDL_ConvertSurface sets blend mode on the created surface.

As such, to fixed flipped sprites, adding

    SDL_SetSurfaceBlendMode(SurfaceFlip, SDL_BLENDMODE_NONE);

right after setting the color key in CGraphic::Flip() suffices to get those sprites on screen.

However, this answer just opens more problems.
At very least, CGraphic::MakeShadow() needs to be rethought, as it seems it won't work (I'm not quite sure what should I be looking at to tell whether or not it does now).

Also, that still doesn't make a difference with the summer problem.
I've got no idea what causes that - as little as opening the map editor with summer tileset triggers this, but neither of the other three does.

galtgendo:
Well, alright, the summer problem was of my own design.

The line in src/map/map_fog.cpp CMap::InitFogOfWar() that originally went:

    s = SDL_DisplayFormat(FogGraphic->Surface);

I've changed to:

    s = FogGraphic->Surface;

but in this particular case (not going into details) it's plain wrong.

It should go

    s = SDL_ConvertSurfaceFormat(FogGraphic->Surface, SDL_PIXELFORMAT_RGB888, 0);

Then things work fine.

That should be the end of phase one.

Some issues remain:
- keyboard input needs to be polished (both wrt. text input and using keysyms)
- there's something odd going on when selecting by drag - rubberband (and mouse) goes into flickering if one of map screen corners is reached (bad description - need to see it yourself)

...there might be be, but can't recall it right now

Andrettin:
@galtgendo that is great!

About MakeShadow(), it is supposed to take a shadow layer file and make it have 50% transparency. In principle, it would be quite useless - since the shadow is a CGraphic rather than a CPlayerColorGraphic, it doesn't need to be indexed to a palette, and thus can already have semi-transparent pixels if so desired. The only problem is: Stratagus can't properly flip images which aren't indexed to a palette; although there is code for that, doing so simply crashes the game (at least with SDL1). So the MakeShadow() function is the only way to get flipped semi-transparent shadows, but it would become completely useless (and IMO should be removed) if the flip issue for images with greater color depth is fixed.

galtgendo:
Lest I forget, there are some minor tweaks that might or might not matter:
- in LoadGraphicPNG(CGraphic *g), moved 'if (ckey != -1)' block after palette is set; not sure if it has any effect, but more technically correct (if there's a palette, transparent color is only meaningful within it)
- in CColor::Parse, 'this->A = 0;' -> 'this->A = 255;'

On a funny note, I've got a bit anal tweak to ConvertPalette in wartool.cpp; the way I see it, it would be more correct if 'pal[i] <<= 2;' went 'pal[i] = (pal[i] << 2 | pal[i] >> 4);'.

On not quite related note: during trying to figure out what was wrong with summer set, I've noticed something interesting: the layout of tiles differs between original data and those shipped with wargus  and not just due to the extra rows.  Unless I'm missing something, the tileset scripts in wargus are either wrong for Warcraft 2 or wrong for Tales.

Andrettin:
Tales has a few unused tiles (the cliffs for instance). Is that what you're referring to?

galtgendo:
For example lets look at winter.png: in wargus version, it has a black tile at index 187, but originally such tile was only present in summer.png. the author of wargus version seem to want to unify the layouts. I don't quite understand the tileset setup script (or lua altogether), but it doesn't look as if it could handle both at the same time.

Andrettin:
Ah yes, you're right. I just looked at the scripts and Wargus does have two different sets of scripts for Tales and WC2 tilesets as of the new version.

galtgendo:
Yes, but both seem to refer to the same file...

Andrettin:
I think the Wargus extractor replaces the Tales files with WC2 ones after extracting. It's something that the Wargus program does rather than Stratagus per se.

galtgendo:
The way I see it, that just means that at minimum both wartool and one of those scripts per tileset need to be fixed to to refer to a different file.

Kintobor:
Tales uses the Wargus summer tileset lua script for all the tilesets.

galtgendo:
So, anyway, just to make sure that I didn't miss anything, attaching another patch of the current state.
It's still against 2.3.0.

[stratagus_2.3.0-sdl2-x4.patch.txt](https://github.com/Wargus/stratagus/files/67474/stratagus_2.3.0-sdl2-x4.patch.txt)

Yet, there's something I didn't seem to stress sufficiently from the start: this patch, despite it's near completeness, was meant to be a start of a discussion, not something ready to commit.
Note all the pseudo-debug stuff that was added (for *my* convenience) and is still to be removed.
Also, in this patch there's still a bit of the stuff that was touched only to figure out one or more of the described problems (like for example, that block in DrawSubClip) that's not meant to be commited at all (thought that set of functions could have used some streamlining regardless)...

galtgendo:
As for the tilesets, what I'm saying is that things can't work right the way they're now:
when wartool will overwrite the png file either the script becomes incorrect, or it already is that way for the new files from wargus.

As a sidenote: identify (from ImageMagick) complains about something being wrong with bKGD value in icons.png (for all four tilesets) in those wargus files. Obviously, it doesn't stop the from getting displayed, bit interesting regardless.

Kintobor:
I don't see why the scripts need to refer to different files. Only one of the scripts are used based on which data set you have installed.

galtgendo:
:sigh:
Lets try making baby steps:

Option 1: current wargus tileset script is correct for original data
If the user considers wargus tileset an improvement and keeps it past wartool, as the layout of data differs, script will become incorrect.

Option 2: current wargus tileset script is correct for wargus data
If the user wants the original tileset, as the layout of data differs, after wartool overwrites data, the script becomes incorrect.

timfel:
@galtgendo: if you use the aleonas tales data, as per https://github.com/Wargus/wargus/blob/master/scripts/tilesets/summer.lua, the tileset script will be loaded from the tales subfolder. if you extract the data, the tileset png will be overwritten, and (because it has a different format), the scripts will now load the tileset script from the wargus subfolder.

galtgendo:
It seems I'm missing something here.

Is it that as soon as wartool runs, the tales data simply gets corrupted ?

Cause while scripts differ, they refer to the same png file...

Also:

    Tales uses the Wargus summer tileset lua script for all the tilesets.

so, what's the point of the other tileset png files ?

Oh, and wrt. MakeShadow(): I wasn't asking "where in the code is it ?", I was asking "where on the screen can it be seen ?", cause either I don't see it, cause the patch breaks it, I do see it, but not consider it a shadow or the effect is just too subtle for me.

Andrettin:
MakeShadow() is not used by Wargus; it is only used if the Stratagus game in question defines shadow sprites separate from the unit's main image files.

timfel:
That is the case in stargus, iirc

timfel:
@galtgendo yes, as soon as you run wartool, the tales data is simply overwritten and the scripts for tales no longer apply.

DinkyDyeAussie:
The whole reason there are different files for each tileset is because each tileset, while being layed out differently, also has a different colour pallette and therefore a different colour cycle for it.

Trust me I'm the one who figured out the different colour cycling numbers for each tileset. Unless we can have the same png layout for each tileset.png and then test for which one we are using for the map and theeeen have the appropriate colourcycle routine invoked, I don't see why we need to "fix" the warcraft 2 tilesets inside wartool.

If it's not broken don't waste time trying to fix it.

If the Aleona's Tales tileset isn't working properly maybe we should work on the files relating to tgat game....

galtgendo:
@timfel you know, stargus right now reliably crashes upon loading a save game.
Not even a specific one: with several retries, starting a new single player game (on the default map - some of the maps crash upon load - probably unimplemented units) then saving, results in a crash on each load attempt.

Also, there's a small bug in scmconvert: in scm.cpp in SaveTrigger, Action 7 needs to be guarded against TriggerNumber == 0 or you'll get a segfault. Not sure what's the correct move there, so just using "(none)" as a string in that case.

Oh, and this might just be a case of me doing something wrong, but bunkers seem to lower fire range instead of increase it (or at least it seems enemy units have bigger range that those in the bunker).

Just rechecked it with sdl1 version - funny thing: while shadows seem indeed broken (though it shouldn't be hard to fix that), the most outstanding problem (missile turrets) is actually already present there.

The UI also needs conversion to build progress within unit info window and those building HP  (?) bars look suspicious. Though, both of those are most likely within lua scripts.

timfel:
Yes, stargus is not very complete or anywhere near well tested.

galtgendo:
PS: what triggers the massive slowdown upon picking repair in SCV ?

I thought it was my patch, but it's already present in SDL1.

galtgendo:
PPS: not really on topic, but do you have any plans of adding a shader that's closer to the real HQX, besides the one mislabeled as such ? There is a version of it on Github, though it has a semi-obvious typo (and I'm still unsure how it relates to the original hqx, as what that used doesn't seem to have been YUV (or anything close to it)).
I'm talking about commit 3a29810853ae90ae1ba1cf1d96b5cca4f8d89e42 - though I'm not sure how raw OpenGL can be mixed with SDL2 properly, as it bypasses SDL_Renderer (which in hardware is already implemented by shaders). What little I've read about it says it's quite tricky.

timfel:
I don't know what you mean by real and mislabeled. I've taken the shaders from some snes emulator, they are vanilla versions pf the 2x hqx, supereagle and so on shaders. As far as I could tell, the algorithms are all correct wrt their pseudo code variants i found on wikipedia, but I could be wrong, of course. However, in any case we wouldnt be the only project to mislabel those shaders if they are indeed mislabeled.

timfel:
There are of course many versions on github that are in pure c++, but I have no interest in adding these, because I prefer such post processing to be done on the GPU.

galtgendo:
See http://filthypants.blogspot.com/2014/06/true-hq2x-shader-comparison-with-xbr.html for details.

As for the shader, I'm talking about https://github.com/Armada651/hqx-shader - it has a glsl version, but it has that typo I've mentioned (the author disagrees, but AFAICT is wrong).

timfel:
The common shaders repo mentioned in that blogpost is the one i have taken some shaders from, though I dont remember if the version of hq2x I took is from there. I also looked at that other repo you mentioned, but something was wrong with it or I didn't use it right (dont recall what), so I didn't use it.

galtgendo:
Well, besides the typo, Armada shader is likely (that YUV catch aside) the closest to original hqx as is available.

The typo is that is does


     vec2 ps = 1.0/TextureSize;
     float dx = ps.x;
     float dy = ps.y;
     vTexCoord[0].zw = ps;

in the vertex shader part, then

    float dx = vTexCoord[0].w;
    float dy = vTexCoord[0].z;

which switches the order of those vars, leading to incorrect results.

galtgendo:
By closest, I mean there seems to be some float rounding problem lurking, that makes the shader output not quite equal to the C version.

hanetzer:
Just a curious thread bump, any progress towards this?

keithstellyes:
I'm curious about this too

Andrettin:
By the way, why is there so much new code that is commented out?

PR #165, opened 2015-12-19
Added thread on sound server

marcelofg55:
-- Sound server now uses a thread to avoid mixing the buffers inside the sdl callback.

Hi, I've added a thread for the sound server, so that the callback doesn't need to process all the mixing from sounds and music.
I've tested it on linux and mac.

Regards,
Marcelo


timfel:
Thanks!

PR #168, opened 2015-12-27
Fullscreen is toggled fine now on OSX

marcelofg55:
Tested on OSX Yosemite.


timfel:
I have no macbook to test this quickly right now, so if you say it works ... ;)

PR #172, opened 2016-01-08
Fixes deadlock and issue on windows with audio system

marcelofg55:
This fixes two issues that was encountered with the new audio thread code while testing with Wyrmsun:
- Fixed deadlock within the sound system
- Fixed issue on windows only that made the audio sluggish


Andrettin:
Thank you!

PR #174, opened 2016-01-12
Refactor target searching for splash missiles to make friendly fire configurable

timfel:
Prompted by https://github.com/Wargus/war1gus/issues/63, I looked at the code for target finding when we have a missile with splash damage. In Warcraft 1, catapults simply will not fire if there is a chance it might hurt friendly units (even if they would survive). Stratagus seems to use a cost-weighing function that accepts friendly fire, and only slightly prefers to attack units that are not near friendly units if possible.

I refactored this code to be slightly more explicit, and I added a FriendlyFire flag (by default set to true) that determines if we want Warcraft 1 type behavior for a unit type. I don't want to merge this without some feedback, so that's why the pull request

### DIFF
--- a/src/include/unittype.h
+++ b/src/include/unittype.h
@@ -170,6 +170,7 @@ enum {
 	NORANDOMPLACING_INDEX,			/// Don't use random frame rotation
 	ORGANIC_INDEX,					/// Organic unit (used for death coil spell)
 	SIDEATTACK_INDEX,
+	NOFRIENDLYFIRE_INDEX,           /// Unit accepts friendly fire for splash attacks
 	NBARALREADYDEFINED
 };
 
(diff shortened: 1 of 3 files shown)


Andrettin:
It looks good, but it's difficult to know without testing in practice.

I would remove the "FriendlyFire" boolean and leave just the boolflag, though. Some other boolflags do this, but it is an unnecessary duplication. 

Doing that would entail removing this line in unittype.h:
bool FriendlyFire : true;

...and removing these lines in script_unittype.cpp:
type.BoolFlag[FRIENDLYFIRE_INDEX].value          = type.FriendlyFire;

[...]


} else if (!strcmp(value, "FriendlyFire")) {
	type->FriendlyFire = LuaToBoolean(l, -1);

...as well as removing this in unittype.cpp:
FriendlyFire(true),

And finally, changing this in unit_find.cpp:
if (sgood > 0 && !attacker->Type->FriendlyFire) {

to:

if (sgood > 0 && !attacker->Type->BoolFlag[FRIENDLYFIRE_INDEX].value) {

Doing these things would mean that the value would default to false. If you want it to default to true, you could change the boolflag to be "NoFriendlyFire" instead.

timfel:
Okay, I've changed this to only use the bool flag, and also squashed it into one commit. If we find that it breaks other things, there will only be this one commit to revert. I propose we merge this

Andrettin:
Great, I agree :)

PR #176, opened 2016-02-02
Fix test for data extraction on non Windows system

ckorn:
It was not able to find the file because of the backslash notation.
Use slash notation on non Windows systems to fix it.


timfel:
Thank you!

PR #177, opened 2016-02-03
Force re-generating the textures for color cycled graphics in OpenGL

### LINKED ISSUE #147: Color Cycling Doesn't Work under OpenGL
Color cycling doesn't work under OpenGL.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2015-06-09T19:22:41Z
 * owner: andre-ng
 * the launchpad url was https://bugs.launchpad.net/bugs/1463532

timfel:
Fixes #147 (and also https://github.com/Andrettin/Wyrmgus/issues/27)

### DIFF
--- a/src/include/video.h
+++ b/src/include/video.h
@@ -439,7 +439,7 @@ extern int LoadGraphicPNG(CGraphic *g);
 #if defined(USE_OPENGL) || defined(USE_GLES)
 
 /// Make an OpenGL texture
-extern void MakeTexture(CGraphic *graphic);
+extern void MakeTexture(CGraphic *graphic, bool force = false);
 /// Make an OpenGL texture of the player color pixels only.
 extern void MakePlayerColorTexture(CPlayerColorGraphic *graphic, int player);
 
--- a/src/include/video.h
+++ b/src/include/video.h
@@ -74,7 +74,8 @@ protected:
 		Width(0), Height(0), NumFrames(1), GraphicWidth(0), GraphicHeight(0),
 		Refs(1), Resized(false)
 #if defined(USE_OPENGL) || defined(USE_GLES)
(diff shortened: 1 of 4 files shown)


timfel:
I realize this implementation is inefficient. The right way (TM) would be to use a shader, an intermediate solution may be to just cache the color cycled textures. As it is right now, it doesn't bog down performance for me too much, so I refrained from optimization.

timfel:
Ok, changed it to lazily initialize all the color cycling textures at once and then just switch between those instead of cycling the palette all the time.

timfel:
@andrettin does this caching work for you?

Andrettin:
Looks good to me :)

I would also delete the color-cycled textures when ClearAllColorCyclingRange() is called, though. This is because a game could use this function to clear color cycling ranges and then redefine them when changing tilesets (if the different tilesets have different color-cycling indexes).

timfel:
@Andrettin but can you call that during the game? If you start a new map with a new tileset, I think the textures are re-created anyway.

Andrettin:
I call it in Wyrmsun with scripts.lua (so just before a game starts). Textures are created and loaded when CclDefineTileset() is used (but they are never recreated once created though). Scripts.lua is loaded right after that; I could put it to be loaded before, though. Either way, I think it is more intuitive and safer to delete the color-cycled textures of the graphics when color-cycling colors are cleared, and then re-create the textures with the new color-cycling ranges after that.

I don't want to give you trouble though, so I could push a commit doing that after you merge this PR, if that's alright.

timfel:
If you know how, go ahead and push it - I'm not protective of branches I open ;) I just don't have time for this right now, I only do housekeeping right now while other work is compiling.

PR #182, opened 2016-02-16
Allow building over decorative buildings without HP

timfel:
This change is needed for War1gus. In War1gus, you should be able to build directly over roads. One way to do this would be to make them decorations but not buildings, but then you can built them on map tiles marked with `no-building`. This solution makes roads be buildings again, but since they do not have HP, they _can_ be built over. If there are no objections to this, I'll merge this later tonight.


Andrettin:
Good idea :) I don't see any side-effects that this could cause, either.

PR #186, opened 2016-03-14
Fixes for screenshot code with opengl

### LINKED ISSUE #156: Menu screenshots appear completely black if shaders are active
If the shaders are active, screenshots made of the menu appear completely black. If playing with the "-x -1" command line, this doesn't happen.

timfel:
This doesn't rely on the internal organization of std::vector as an array and accounts for shader support as well as possibly different Viewport sizes vs Video sizes. Should fix #156 and might help with https://github.com/Andrettin/Wyrmgus/issues/73

### DIFF
--- a/src/include/shaders.h
+++ b/src/include/shaders.h
@@ -9,7 +9,7 @@ extern PFNGLBINDFRAMEBUFFEREXTPROC glBindFramebuffer;
 #endif
 extern GLuint fullscreenFramebuffer;
 extern unsigned ShaderIndex;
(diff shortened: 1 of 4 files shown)


timfel:
@Andrettin couldn't sleep, so here goes one tired try ;)

Andrettin:
Awesome :)

timfel:
This now also fixes the issue where a screen might appear 'upside down' with opengl if shaders are supported by the opengl library, but the hardware will not accept the shader program.

PR #189, opened 2016-05-26
Enable the use of libretro shaders

timfel:
Rather than having some few compiled in shaders, this patch allows shipping the single pass libretro/common-shaders after converting them to glsl (which is done automatically using a script). This allows users to make or add their own shaders and it allows us to ship some nice shaders by default (I've included different pixel-art scalers, different smoothing algorithms, artsy stuff like embossing or making it look like an old CRT or phosphor TV).

The main change is loading shaders from the shaders directory in the game folder. Thus, each game can decide which shaders to ship depending on which look good with the art in the game.

I especially like the CRT one for Warcraft 1:
image


timfel:
@Andrettin you were wondering about shaders for zooming into parts of the screen - this approach might be useful here. All we need is a way to select the shader from lua, and then you just ship shaders for different zooms.

Andrettin:
That's interesting. Is it capable of doing that only when rendering units/etc.? The map area in Wyrmsun is overlayed by some semi-transparent pixels of the UI, so just getting a square from the screen unfortunately wouldn't work in that case.

timfel:
@Andrettin Oh, I wasn't aware of that. While it is possible, it requires a little more thought on how to structure things. Right now this is just a fullscreen thing and mostly for looks.

PR #201, opened 2017-02-17
Fixed FSF address

susnux:
Fixed FSF address in documentation.
See also [gnu.org](https://www.gnu.org/licenses/old-licenses/gpl-2.0#SEC4).

Without this fix this leads into a lot of RPMLint warnings when building for openSUSE.


timfel:
Thanks!

PR #216, opened 2017-10-28
Windows 7 crash, added an option to CMake :

ChefKeeper:
If stdio redirection is enabled compile with WIN32 (windows app with no console)
else when no redirection we need a console app to write all the stdio

for : https://github.com/Wargus/wargus/issues/255


pali:
That is not a fix. It just disables code and disables WIN32 GUI subsystem.

timfel:
I cherry picked these to mastet

PR #226, opened 2018-06-11
Don't use AIForce's GoalPos after calling ReturnToHome()

DanielHartl:
I get an AccessViolation without this fix. It seems to me that it doesn't make sense to use force.GoalPos after force.ReturnToHome has been called since it's set to -1/-1


timfel:
Thanks!

PR #231, opened 2018-06-26
Adding TCP for multiplayer games

DanielHartl:
UDP doesn't work or is hard to set up (if both parties are behind a NAT router). TCP support makes things much easier if there is a dedicated server with a public ip.

Caveat: this change breaks publishing games to metaserver. I think we should redefine the role of metaserver. Additionally to just enable discovery of newly hosted games it should actually start the dedicated server instances and register them with itself to allow clients to discover and connect to it.

### DIFF
--- a/src/network/netconnect.cpp
+++ b/src/network/netconnect.cpp
@@ -193,7 +193,11 @@ void CClient::Init(const std::string &name, CServerSetup *serverSetup, CServerSe
 }
 
 void CClient::Open(bool udp) {
-	if(udp) {
+	if (_clientConnectionHandler) {
+		Close();
+	}
+
+	if (udp) {
 		_clientConnectionHandler = new CUDPClientConnectionHandler(this->serverHost);
 	}
 	else {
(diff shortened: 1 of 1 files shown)


src/network/net_connection_handler.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+int CTCPServerConnectionHandler::Recv(unsigned char* buf, int len, CHost* hostFrom) {
+	// round robin across _clientSockets to avoid chatty client flooding client/server comm
+
+	int skip = _skipClient;
+	int take = _clientConnections.size();
+
+	_skipClient = (_skipClient + 1) % _clientConnections.size();
+
+	for (auto& it : _clientConnections) {
```
DanielHartl: need to close and remove connections if the underlying socket has been closed

timfel:
Hi,
thanks for taking a look at this! Yes, the metaserver should try to make sure we can connect. The current approach tries to use NAT tunnelling through the metaserver to connect, but this often does not work. Indeed, a dedicated server running on the metaserver on a public IP/Port would make things easier. Feel free to experiment a bit and ping @mimooh, @Andrettin or me if you want to discuss a general overhaul of the networking code. I believe Wyrmgus didn't change too much in that area, either, so a more modern implementation of networked games would maybe benefit them too?

PR #240, opened 2018-12-09
Adaptation to Lua 5.3

Nassiel:
I created the code to adapt stratagus to lua 5.2 and 5.3. While it should keep the compatibility with 5.1 I cannot test it, so I ask you to test it before in your machines.


timfel:
Hi, thank you, I will test it with lua 5.1 as soon as I have time. I don't want to loose compatibility, because I tried moving to 5.2 and 5.3 before, and they are much slower than 5.1 in Stratagus

PR #250, opened 2019-06-21
enable autocast during attack-move

talas:
https://github.com/Wargus/wargus/issues/175

Allows autocast to trigger during an attack-move.
(I assume this to mean anytime the unit has the attack order, but not actually attacking something.)
Perhaps it must be configurable?
And should it also trigger while attacking?
Or should it not trigger when you target attack a unit?

### DIFF
--- a/src/action/action_attack.cpp
+++ b/src/action/action_attack.cpp
@@ -605,6 +605,11 @@ void COrder_Attack::AttackTarget(CUnit &unit)
 		unit.Waiting = 0;
 	}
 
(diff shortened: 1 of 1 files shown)


timfel:
I think it makes sense never to auto cast if you're actually attacking something. We might want to be able to configure this, because I wouldn't want e.g. an ogre mage on attack-move start to auto cast bloodlust on the retreating units it meets. But we can fine tune later, this is great.

PR #251, opened 2019-06-26
let EditorChangeSurrounding recurse until done

talas:
Fixes tiling issue when placing forest in dark water and similar situations involving dark water, dark grass or forests being placed on water.
May cause some organic growth of coasts.
It's possible that the real problem exists somewhere below tileFromQuad, but I couldn't figure it out.

To reproduce the issue:
Make a large body of dark water.
Try placing a forest tile in the middle.

On my computer, this results in a cascading tiling error.
Only tested with wargus (BNE version on Linux).

### DIFF
--- a/src/editor/edmap.cpp
+++ b/src/editor/edmap.cpp
@@ -44,8 +44,8 @@
 #include "unittype.h"
 
 
-/// Callback for changed tile (with direction mask)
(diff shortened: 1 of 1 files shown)


timfel:
Nice, thanks!

PR #252, opened 2019-06-29
Colorcycling change

mxdamien:
Slowed down color cycling to better mimic original Warcraft 2 speed.


timfel:
Why close it again?

PR #254, opened 2019-06-30
add minimal support for attract/demo mode

talas:
in wargus, this means watching 2 computers fight eachother.
any interaction will end the demo.

Could also create a IsDemoMode() helper instead of the "weird?" (GameRunning && ThisPlayer->Type == PlayerNobody) check.

The term demo is a bit confusing here though, as it has nothing to do with the shareware version.

### DIFF
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -903,6 +903,15 @@ void CreateGame(const std::string &filename, CMap *map)
 		}
 		CreatePlayer(playertype);
 	}
+	if (!ThisPlayer && !IsNetworkGame()) {
+		// In demo or kiosk mode, pick first empty slot
+		for (int i = 0; i < PlayerMax; ++i) {
+			if (Players[i].Type == PlayerNobody) {
+				ThisPlayer = &Players[i];
+				break;
+			}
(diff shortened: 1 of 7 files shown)


talas:
To test the changes, you can use this in wargus:
https://github.com/Wargus/wargus/pull/306

timfel:
Sorry, I dropped the ball on this. This is cool stuff! I agree, if you could create a helper function isDemoMode or such, it would make the code easier to read. Could you do that?

timfel:
I will merge the wargus change after this one, because the way the build works master binaries are created immediately with the Stratagus master, so this here has to be merged first.

timfel:
Also, if you're interested in working some more on Wargus and Stratagus, I'd be happy to add you to the dev team, just let me know. You can also ping me on Gitter of you want to discuss plans or changes more directly

talas:
Thanks for the offer, I'll consider it.
Note that I'm currently more interested in making a rebalanced/modernized esports wargus than original gameplay. Which sounds like fork material I guess. But, I'll be happy to upstream anything that makes sense.

PR #255, opened 2019-07-07
add a simple explore action

talas:
flies around randomly, much room for improvement

You can test it in wargus by adding a button for it, like so in scripts/human/buttons.lua:
DefineButton( { Pos = 3, Level = 0, Icon = "icon-human-patrol-naval",
  Action = "explore",
  Key = "x", Hint = _("E~!XPLORE"), Popup = "popup-human-commands",
  ForUnit = {"unit-balloon"} }

Not sure what is supposed to be written in the copyright section. Let me know if I messed up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,7 @@ set(action_SRCS
 	src/action/action_built.cpp
 	src/action/action_defend.cpp
 	src/action/action_die.cpp
+	src/action/action_explore.cpp
 	src/action/action_follow.cpp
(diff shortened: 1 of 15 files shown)


talas:
Oh, and this needs to be tested in multiplayer.

timfel:
This is actually a cool action for the ai scripts also - the ai doesn't need it, of course, but in the original flying machines fly around which gives or a more natural feeling

timfel:
As for the button, this may be a good candidate for an "auto" action on the move button for recon units

PR #256, opened 2019-07-14
Various small ui/gfx/sound fixes

talas:
* Don't let pointtopoint missiles wobble
* Don't spam "Game Resumed" from SDL
^ On my laptop, I kept getting "Game Resumed" popping up on the status bar.
* Hack to make mana visible for units on bottom row
* Play transport docking sound more reliably
* Play random building noises
* Don't show wood being cut under fog of war
* Fix wood not updated by ally vision
* Fix "revealer unit" visible on minimap when casting holy vision in wargus.
* Fix missile not targeting center of 2x2 and 4x4 units.

### DIFF
--- a/src/action/action_built.cpp
+++ b/src/action/action_built.cpp
@@ -267,6 +267,11 @@ static void Finish(COrder_Built &order, CUnit &unit)
 
 	const int maxProgress = type.Stats[unit.Player->Index].Costs[TimeCost] * 600;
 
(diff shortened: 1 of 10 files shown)


timfel:
Not sure about the "Game resumed" thing - I only get this in windowed mode when the mouse leaves/re-enters the window or when alt+tabbing. The other fixes look good, though.

PR #257, opened 2019-07-15
add flag for unittypes that should snipe from distance

talas:
Can be used as a partial fix for https://github.com/Wargus/wargus/issues/177 (first issue: ballista v.s. cannon tower).

If enabled on ballista/catapult, this will make them prefer not going closer to the target than they have to.
It's not perfect, if you have 4 ballistaee one might go too close still.
Another option could be to increase cost of all "threathened" tiles, but I don't think the original did this.


timfel:
I'm wondering if there's a reason we need that flag? Wouldn't we want all units to shoot from the distance if at all possible?

talas:
If you enable it for archers, you can see that they spend some extra time when pathing towards a target.
I think you generally won't benefit from this flag, unless you are attacking something that can't move (towers) and that you outrange.

timfel:
Ok makes sense.

talas:
Well currently it just sets the minimum range for the pathfinder, so that the unit will be more likely to shoot from maximum range.

If we had something like a horse-archer, then that should probably try to maintain the distance as well.. but currently theres no unit in wargus that can both move fast and shoot far.. So I didn't try add that.

I'm thinking the ballista/catapult and battleships could benefit from this change.

timfel:
@talas are you planning to open a PR to wargus to add these to ballistas and battleships?

timfel:
(I'll do the one for war1gus catapults, unless someine beats me to it)

PR #258, opened 2019-07-16
fixes for terrain-harvester and maybe lose-resources

talas:
Previously, it was possible to tell a unit to return goods with the terrain-harvester trait before it reached full capacity.
The effect was that the terrain was left intact and the worker would return what it had mined so far.
Thus you could get unlimited amount of resources from the same terrain tile (we don't track resources in terrain).

Also partially implemented the 'lose-resources' trait. I'm not exactly sure how it's supposed to work, but now kinda does what it says it should do in comments/doc.
Perhaps it should also be triggered in COrder_Resource::LoseResource ..?
But, then we might need to develop the terrain-harvester some more first. (workers cut the same tree in wargus, in original they try hard to spread out).

### DIFF
--- a/src/action/action_resource.cpp
+++ b/src/action/action_resource.cpp
@@ -167,6 +167,15 @@ static bool FindNearestReachableTerrainType(int movemask, int resmask, int range
 	order->CurrentResource = harvester.CurrentResource;
 	order->DoneHarvesting = true;
 
+	if (harvester.CurrentResource) {
+		const ResourceInfo &resinfo = *harvester.Type->ResInfo[harvester.CurrentResource];
+		if (resinfo.TerrainHarvester && harvester.ResourcesHeld < resinfo.ResourceCapacity) {
+			// Need to finish harvesting first!
+			delete order;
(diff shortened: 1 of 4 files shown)


timfel:
This means now I can no longer send a unit back with less than full capacity from trees, correct? If I'm reading that right, I think the real fix must be to track resources in the terrain.

timfel:
Ok, I checked on the original and it is indeed the same, you cannot cancel and multiple peons chopping the same tree leads to one starting over on the next tree.

timfel:
The lose-resource trait should probably also actually work, and as you say the units should spread out more by default. But we can fix that in a separate PR

talas:
Hehe, I do agree that being able to return wood early is a cool feature.
Do we want wood cutting to be 1:1 with original?
I think the current system (even with this patch) means you can gather wood faster on average.
(since workers can't make the mistake of cutting the same tree and having to start over).

timfel:
Indeed, this one is faster. I think it's fine to leave it, we don't have to match the original 1:1 here.

PR #260, opened 2019-07-16
check dependencies for ai when upgrading

talas:
Previously the AI would cheat if instructed to upgrade city hall before it could with AiUpgradeTo.
Land/Sea/Air AI still works with this patch at least.. didn't test campaign.


timfel:
I'd rather have this behind a flag (I think there's AI scripts in both Wargus and War1gus that ignore the dependencies)

timfel:
Finally merged this. Thanks @talas :)

Issue #263, opened 2019-11-28
Stratagus starting and closing on playing video if OpenGL enabled


ilwar2:
Trying on win7:
Just builе last version from github, proceeded new clean war2 resources, moved war2 resources instead of aleona tales resources into c:\wargus, erased AppData\Roaming\Stratagus\, started wargus, everything works perfectly.
After that i switched opengl on, exited, and... stratagus have not been started anymore: window border appears of a second and disappears. No crash, just closing.

AppData\Roaming\Stratagus\stdout.txt  contains standard message, stderr is empty.

Compile options ZLIB BZ2LIB VORBIS THEORA FLUIDSYNTH MNG OPENGL WIN32 
Tried stable version 2.4.2 (your build) - same effect

Would be great if you can make some additional logging to opengl part for debug purposes.

ilwar2:
Got it!

After I removing a directory "videos", it works fine with opengl.
So, looks like it have been closed after trying to playback video in opengl mode.

Debug logs and fixing the problem are still issues.

WargusCI:
Hi,

why do you need to delete the AppData folder? Does it not work when run
from there? If so, that's another bug. You should not have to touch the
installation folder at all.

ilwar2:
I am deleting data from the installation folder (I mean c:\wargus by default) because:
after I unpack the resources and try to run wargus.exe - it crashes.

That is no problem for me as I can run stratagus.exe, it works, but uses data from the current folder.
I don't need aleona at all, so I remove all the data and move the appdata to c:\wargus.
Everything works fine till I switch opengl on.

ilwar2:
I am deleting data from the installation folder (I mean "c:\wargus" by default) because:
after I unpack the resources and try to run wargus.exe - it crashes.

That is no problem for me as I can run stratagus.exe, it works, but uses data from the current folder.
I don't need aleona at all, so I remove all the data and move the appdata to c:\wargus.
Everything works fine till I switch opengl on.

Upd: 
Now I installed wargus again to different directory, unpacked data from BNE version, it have been unpacked to appdata\stratagus, after that I copied data overwriting aleona,
And now this version have different stratagus.exe: version 2.4.2.0, file size: 2121728, 12.06.2017 12:43
It works fine with videos and opengl.

And when i just install clean wargus 2.4.2 from your github, its stratagus.exe: version 2.4.2.0, file size: 2122752, 29.09.2018 22:03.
This exe closes right after it tries to playback logo.ogv from videos.
[...]

PR #264, opened 2019-12-02
opengl crash on video playback fixed

ilwar2:
fix for this issue: https://github.com/Wargus/stratagus/issues/263

### DIFF
--- a/src/video/sdl.cpp
+++ b/src/video/sdl.cpp
@@ -197,6 +197,11 @@ static bool IsExtensionSupported(const char *extension)
 	len = strlen(extension);
 	start = extensions;
 	while (true) {
+		if (!start)
+		{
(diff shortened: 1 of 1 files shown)


timfel:
Thanks! the `fprintf` doesn't compile on Linux due to a format difference, in any case I think we can just omit the debug output. Can you remove the line?

ilwar2:
Of course, that's up to you! Let's omit it.
I'd prefer to have the logs about everything strange happening in the project (for example i have no idea why glGetString(GL_EXTENSIONS) fails sometimes, it actually shouldn't), but i don't know if you use any function or macro for such debug purposes.

Updated, build fails on LUA51 some way, at least that's not my fault now...

timfel:
Yeah, that's a change in the homebrew env, we'll do a separate PR to fix that.

PR #266, opened 2019-12-20
Fix condition check

ipochto:
`if (ifa->ifa_flags & IFF_UP == 0) continue;`
== has higher priority than &


timfel:
Sorry, should I still merge this?

ipochto:
I think so, because condition check is still wrong )
But I not sure if this part of code will execute at some time

PR #267, opened 2020-02-13
Increase buffersize due to possible overflow

mxdamien:
Hey guys.

Using gcc version 7.4.0 I get a warning in CInfoPanel::Draw() in src/ui/mainscr.cpp when building in release configuration. Anybody else having that problem?

_stratagus/src/ui/mainscr.cpp:1168:6: error: ‘%lu’ directive writing between 1 and 20 bytes into a region of size 5 [-Werror=format-overflow=]_

Increasing the buffersize to 22 solves the problem for me.

Regards

David


timfel:
Thanks!

PR #271, opened 2020-04-04
Automatic Targeting.

ipochto:
There were some problems with automatic target selection. 
- After receiving attack-move order attacker chose a target and may fixing on it, even if a higher priority target is appeared nearby. F.e. attacker may beat barracks and don't switch to the appeared from it new enemy unit, or may attack farms, but do not switch to peons nearby.
- Unit being under attack by tower, may attack farm, and do not switch target, as result - die.
- etc.

Maybe that was not big problem for single player and campaign, but for multiplayer game that was terrible. 

What was done. 
1. Was made some fixes in original code, now units don't fixed on once selected target when in auto attack mode. Made some refactoring.
2. As an option was implemented [starcraft 2 automatic targeting algorithm](https://liquipedia.net/starcraft2/Automatic_Targeting). With some simplifications of course *(f.e. we don't have units with multiple weapons, closest angle also don't calculating (yet))*, but mainly it was represented. It is simpler than original *(which counts such things as posible splash damage to ally units or if target can be killed by a single shot etc.)*, but IMHO better fit for multiplayer. Because of that it named SimplifiedAutoTargeting.
Target priority briefly:
    * Once a target is chosen, it will remain the target until it ceases being a valid target *(e.g., dies, becomes invisible)*, leaves the `AttackRange` or a higher-priority target enters the `AttackRange`.
    * A unit (or building) considers all valid targets within it's the `ReactionRange`.
    * The criteria for target selection are, from most to least important:
    -- Which targets are threats to me? *(If it can attack me - it's threat, in planes to add optional parameter `AlwaysThreat` to units definition)*
    -- Which targets have the highest `Priority` values *(defined in units lua-files)*?
    -- Only if I lost *(e.g., dies, becomes invisible, leaves the `AttackRange`)* my previous target: Which target is the closest and has less % of remaining HP? 'Closest' now mean not direct distance from attacker to target, but path length calculated by path finder, to position where weapon's attack range reaches the target. 
 
3. Added new option `Preference.SimplifiedAutoTargeting` to activate it (switch between original and new "Simplified" algorithm). Also, this was added in Wargus menu lua-script, so now you will be able to activate it from the game menu (For single player game by now. Multiplayer in the ToDo list). PR in wargus: https://github.com/Wargus/wargus/pull/319.
4. Units with `MinAttackRange > 0` now automatically will move to better position if their targets is closer than MinAttackRange (of course if there is no any valid targets in range, or it not in `StandGround` state). 
5. I have to add new variable in the `ActionAttack`, which must be (and now it is) saved in the save-file. So new save-files can't be opened with older version of Stratagus.
6. Units with stand ground command no more play attack animation if their target is closer than `MinAttackRange`.
7. Now for attack-move order it draws two lines - one to current automatically selected target, and the second to the ordered attack-move tile.
8. Some small fixes.

### DIFF
--- a/src/unit/unit.cpp
+++ b/src/unit/unit.cpp
@@ -2534,6 +2534,7 @@ DebugPrint("Target: %s[%d], Distance: %d, PathLength: %d \n" _C_ dest->Type->Ide
 	int priority = 0;
 
 	// is Threat?
+	// FIXME: Add alwaysThreat property to CUnitType
 	// Unit can attack back.
 	if (CanTarget(dtype, type)) {
 		priority |= AT_THREAT_FACTOR;
@@ -2607,6 +2608,26 @@ bool InAttackRange(const CUnit &unit, const CUnit &target)
 	return (minRange <= distance && distance <= range)
 			&& (!GameSettings.Inside 
 				|| CheckObstaclesBetweenTiles(unit.tilePos, target.tilePos, MapFieldRocks | MapFieldForest));
+}
+
+/**
+**  Returns true, if tile is in attack range of the unit and there is no obstacles between them (when inside caves)
+**  @todo: Do we have to check range from unit.Container pos if unit is bunkered or in transport?
+**   
+**  @param unit    Unit to check for.
+**	@param pos     Checked position.
+**
+**  @return       True if in attack range, false otherwise.
+*/
+bool InAttackRange(const CUnit &unit, const Vec2i &tilePos)
+{
+	const int range 	= unit.Stats->Variables[ATTACKRANGE_INDEX].Max;
+	const int minRange 	= unit.Type->MinAttackRange;
+	const int distance 	= unit.MapDistanceTo(tilePos);	
+	
+	return (minRange <= distance && distance <= range)
+			&& (!GameSettings.Inside 
+				|| CheckObstaclesBetweenTiles(unit.tilePos, tilePos, MapFieldRocks | MapFieldForest));
 }
 
 static void HitUnit_LastAttack(const CUnit *attacker, CUnit &target)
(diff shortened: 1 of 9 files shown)


timfel:
Code looks good! I will try it out locally asap :-)

ipochto:
Close for a while. Found issue. Will reopen after fix.

ipochto:
Fixed. Reopen now.
There was problem with attacking units (except buildings) when they are closer than MinAttackRange.

src/unit/unit.cpp
```diff
@@ -2671,14 +2862,17 @@ static void HitUnit_AttackBack(CUnit &attacker, CUnit &target)
...
-		if (target.CurrentAction() == UnitActionAttack) {
-			COrder_Attack &order = dynamic_cast<COrder_Attack &>(*target.CurrentOrder());
-			if (order.IsWeakTargetSelected() == false) {
-				return;
-			}
-		} else {
-			return;
-		}
+		return;
+		/*
```
timfel, line 2866: This is commented out because the unit will now automatically select the right target, and we don't need to react to this event anymore, right?

ipochto: In general yes, but I supposed there are some rare and certain situations when we will have to react to this event (f.e. there is no targets/or only non-threat target in visible area, someone under fog of war attack this unit). Because of that this code was just commented instead of removing. But as all worked fine I just forgot about this part.
Added to my ToDo list.

ipochto:
Close for a while. Updating `HitUnit_AttackBack()` stuff.

ipochto:
Sorry for delay, but it's appeared to be a bit difficulty than it's seems before.
I think it's ready now.

Ooops, can't reopen because of force-pushed. Made another PR https://github.com/Wargus/stratagus/pull/280

PR #283, opened 2020-05-24
Fixup for wall attack.

ipochto:
There was an issue - a unit which attacks wall will attack it and after door destroyed. There wasn't a condition check for wall destroyed. This PR fix this.


ipochto:
Sorry, moved to separate branch. Reopen.

PR #289, opened 2020-06-11
Add units revelation for players, which had lost their last Town Hall.

ipochto:
Units of revealed players will appear in the minimap. 
Mostly need in the end of game, in cases when enemy units spread out across map and you have to find all of them to finish this game.

If enabled, there is two options "Reveal all units" (set by default) and "Reveal only buildings" (added for future use). Which can be selected in the stratagus.lua by calling `SetRevelationType()`.
Loosing last Town Hall triggered timer, an if player will not start build another TH he will be revealed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: cpp
-os:
-  - linux
-  - osx
+jobs:
+  include:
+    - os: linux
+      dist: bionic
+    - os: osx
 addons:
   apt:
     packages:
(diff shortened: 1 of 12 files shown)


timfel:
@ipochto this looks good to me, is it ready to merge in your opinion?

ipochto:
Yep, I've tested it. There was an idea to automatically chose correct revelation type (all units or only buildings) accordingly to selected end-game-trigger (till last unit or last building), but when I dig deeper... it seems we need to fix game types and corresponded end-game-triggers first (for multiplayer games too). So It will be another PR.

So, yes, I thinks it's ready,  as long as it resolve present problems with finding spread out AI units in the end of the singleplayer games, and now there is option to select desired revelation type for modding.

ipochto:
There is too corresponded PRs - war1gus and wargus. Don't forget about them too. It won't work without them.

PR #290, opened 2020-06-12
Handle SDL2 window resize events.

akce:
This allows using maximum screen space.

I've also assigned the changed dimensions to WindowHeight & WindowWidth, but I do wonder whether those two member vars are needed at all. But that'd be a separate issue..

### DIFF
--- a/src/video/sdl.cpp
+++ b/src/video/sdl.cpp
@@ -539,6 +539,11 @@ static void SdlDoEvent(const EventCallback &callbacks, SDL_Event &event)
 				}
 				}
 				break;
+				case SDL_WINDOWEVENT_RESIZED:
+					Video.WindowWidth = Video.Width = event.window.data1;
+					Video.WindowHeight = Video.Height = event.window.data2;
+					Video.ResizeScreen(Video.Width, Video.Height);
+				break;
 			}
 			break;
 


ipochto:
But this will broke aspect ratio, isn't it?

akce:
> But this will break aspect ratio, isn't it?

no, this brings the SDL2 implementation back to what I see with the old SD1 version.

ie, pixel for pixel, it draws enough to fill avail space rather than stretching etc.

akce:
Actually, i'm slightly mistaken. Aspect ratio will be out for title screen and the graphic behind the main menu.

But this was the case for the SDL1 version too.

The gameplay screen is what i was referring to in my first reply.

timfel:
I think we explicitly want to avoid this behaviour. The chosen rendering resolution should be used, regardless of the window size. Otherwise the viewport size and aspect ratio changes when resizing the window. If someone wants a widescreen resolution, for example, they should pick it in the options, not expect the window resize to change the visible map area.

timfel:
That's also the reason why `Video.WindowWidth` and `Video.Width` are not the same thing - one is the window size, the other the rendered size

timfel:
For example, I have a 4k display here, and I want Wargus to run in a window large enough to see, but if the Window size and rendered size is the same, the units are still going to be tiny. If I want them tiny, I have to select a rendering resolution that makes them so. The window size is purely a means to scale the output.

akce:
Ahh I see and take your point completely.

This proposed change, if used at all, could be part of an "adjust rendering resolution to window size" feature, but is unsuitable as is.

Thanks very much for taking the time to explain.

akce:
I hope i'm not breaching ettiquette, but i'll close this request.

If someone wants an adjust-render-resolution-to-window-size, then that'd be a separate issue.

timfel:
@akce thank you :) if you're interested in implementing it as a commandline option, for example, I would be happy to include such a feature. We can also open an issue to not forget about this

akce:
Thanks @timfel that's kind of you - I'm game so let's give it a try. :)

I'll create an issue/feature-request for it now.

PR #300, opened 2020-08-18
Fixed Access Violation on Windows 10 (x64)

boskee:
hFile should be defined as intptr_t to resolve 0xC0000005: Access Violation in ntdll.dll, as per https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/findnext-functions?redirectedfrom=MSDN&view=vs-2019


timfel:
Thanks!

PR #317, opened 2021-02-27
Adding two commands in CclOrderUnit

SimoneStarace:
I was playing around with the <u>Wargus editor</u> and I found a specific Lua script where I can specify some orders to give to the units inside the <u>.sms</u> files. So I was searching for the commands **stop** and **stand-ground** but every time I tried to give those orders I got this error:

`/build/stratagus-A5HbDh/stratagus-3.0.0/src/unit/script_unit.cpp:870: CclOrderUnit: Unsupported order: stand-ground`

So I have found the commands inside the **commands.cpp** file and I simply added these two commands in the **CclOrderUnit** static function.


timfel:
Thank you!

PR #320, opened 2021-03-16
Add cmake-variants.yaml to allow vscode to build Stratagus according …

ipochto:
…to variants defined in the CMakeLists.txt


ipochto:
image

Issue #322, opened 2021-03-18
Assertion failed when trying to create new action to harvest just removed tree.


ipochto:
Debug version with assertion enabled.

Singleplayer
Map: `maps/skirmish/multiplayer/(8)farscape2008.sms.gz`
8 players (1 player, 7 comps with AI "Land Attack"), 2 teams (4 vs 4), Top vs Bottom, Resources "Quick start" 

After about 1 minute of game starts it crashes with:
`Assertion failed at /home/.../stratagus/src/action/action_resource.cpp:122: FindNearestReachableTerrainType: Map.Field(startPos)->CheckMask(resmask)`

Stack trace:
![image](https://user-images.githubusercontent.com/5509477/111636512-9e7ce300-8809-11eb-8786-7646667a2701.png)

The piece of code that causes the error:
```
if (harvester.CurrentResource) {
        const ResourceInfo &resinfo = *harvester.Type->ResInfo[harvester.CurrentResource];
        if (resinfo.TerrainHarvester && harvester.ResourcesHeld < resinfo.ResourceCapacity) {
            // Need to finish harvesting first!
            delete order;
            return NewActionResource(harvester, harvester.tilePos);
        }
    }
```

It was added by this commit https://github.com/Wargus/stratagus/commit/29f3a34

Tim's comments:
>It's just about the single tile
>Not the entire forest
>The worker starts on one tree, returns, and tries to go back to finish that same tree. But if it's gone in the meantime...
>I think it used to be that it did a completely new search for wood

P.S. Created not to get lost.

ipochto:
![image](https://user-images.githubusercontent.com/5509477/111645453-c3755400-8811-11eb-8c56-64f5dd20d3d9.png)
It seems that he is not just out from a depot and trying to get the same tree. It has only 72 gathered resources, and the tree is disappeared, so another unit harvest it tile.

ipochto:
Fixed in master: https://github.com/Wargus/stratagus/pull/323

PR #323, opened 2021-03-18
Fix/resources gathering

ipochto:
- Removed wrong assertion. Fixes this: https://github.com/Wargus/stratagus/issues/322
- When returning from a depot to gather resources, use the same mine as before if it is still alive; otherwise, search for a resource unit as close to the depot as possible
- Enabled 'H'-click for resources gathering (when click on the empty tile unit will find nearest to this pos resource).
- Enabled 'Harvest" rally-point for Town Halls: now newly produced peasant will go to gather nearest to the rally-point resources. Before, it was possible only for mines.

### DIFF
--- a/src/ui/mouse.cpp
+++ b/src/ui/mouse.cpp
@@ -1315,7 +1315,7 @@ static int SendResource(const Vec2i &pos)
 					if (unit.Type->ResInfo[res]
 						&& unit.Type->ResInfo[res]->TerrainHarvester
 						&& mf.playerInfo.IsExplored(*unit.Player)
-						&& mf.IsTerrainResourceOnMap(res)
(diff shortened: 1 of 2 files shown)


src/ui/mouse.cpp
```diff
@@ -1315,7 +1315,7 @@ static int SendResource(const Vec2i &pos)
 					if (unit.Type->ResInfo[res]
 						&& unit.Type->ResInfo[res]->TerrainHarvester
 						&& mf.playerInfo.IsExplored(*unit.Player)
-						&& mf.IsTerrainResourceOnMap(res)
+					  /*&& mf.IsTerrainResourceOnMap(res)*/
```
timfel, line 1318: A code comment here might be a good idea

ipochto: Fixed

PR #328, opened 2021-04-17
Make it a preference if allies can return goods to each others bases

timfel:
… off by default

### DIFF
--- a/src/include/net_message.h
+++ b/src/include/net_message.h
@@ -31,6 +31,8 @@
 
 //@{
 
+#include "settings.h"
+#include "stratagus.h"
 #include <stdint.h>
 #include <vector>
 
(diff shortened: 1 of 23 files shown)


timfel:
@ipochto what do you think?

ipochto:
Sorry, was out of PC.
I like it. This issue was very annoying because:
1. Allies workers had  been "stealing" gold from your mine. Yes, I know - they are allies, but this is **my** gold, and I spend my money for this town hall )) If you really has no other options - then build a town hall nearby at least.
2. A lot of workers in the single resource gathering route, they create a traffic jam.

One more thing:
This is one of those preferences which will cause desync in the network games (along with auto targeting and a couple of others). I think we should to prohibit to change them for the network games (I mean then game is already started) and check if they mach before game starts, or send their values from host to all clients, or set them in the game creating lobby (like as we do it for fog of war and map revelation). But not sure last variant is optimal.

ipochto:
Moreover I would leave the possibility to change this preference only for certain game mode instead of adding the option to change from the menu.

timfel:
> This is one of those preferences which will cause desync in the network games

Right, I forgot about that - thanks!

src/include/net_message.h
```diff
@@ -95,6 +96,17 @@ class CServerSetup
...
+	union {
+		uint8_t BitField; // IMPORTANT! must be the same size as the struct below
+		struct Fields {
+			uint8_t AiChecksDependencies : 1;
+			uint8_t AiExplores : 1;
+			uint8_t AllyDepositsAllowed : 1;
+			uint8_t SimplifiedAutoTargeting : 1;
+		} Fields;
+		static_assert(sizeof(BitField) == sizeof(struct Fields), "bitfield and storage of shared prefs");
+	} SharedPreferences;
```
ipochto, line 109: Don't you think it's very hardcoded for war1gus and wargus? I mean - stratagus - it's an engine. Maybe to do it in some more general way? 
For example declare in the lua config files which preferences could cause the desync and sync them before game starts.

ipochto: For example "allow multiple town halls" is actual only for war1gus, but has to be synced here too.

timfel:
Finally cherry picked to master

PR #342, opened 2021-05-25
Added more examples of function calls.

SimoneStarace:


src/ai/script_ai.cpp
```diff
@@ -582,11 +596,21 @@ static int CclAiDebugPlayer(lua_State *l)
...
 **
 **  @param l  Lua state.
 **
 **  @return   Number of return values
+**
+** Example:
+**
+** <div class="example">
+** 	<code>
+**		unit_needed = AiNeed("unit-human-barracks")
```
timfel, line 611: Could you make the example a bit longer? First I was confused as I read this as "the function returns the new unit". But it doesn't, right? It just returns if the unit is needed or it's already existing?

src/ai/script_ai.cpp
```diff
@@ -598,11 +622,21 @@ static int CclAiNeed(lua_State *l)
...
 **
 **  @param l  Lua state
 **
 **  @return   Number of return values
+**
+** Example:
+**
+** <div class="example">
+** 	<code>
+**		peasants = AiSet("unit-peasant",3)
```
timfel, line 637: Same here, I think the code example could lead to confusion.

src/ai/script_ai.cpp
```diff
@@ -855,9 +921,20 @@ static int CclAiCheckForce(lua_State *l)
...
+**
 **  Wait for a force ready.
 **
 **  @param l  Lua state.
+**
+** Example:
+**
+** <div class="example">
+** 	<code>
+**		-- Wait the force number 1
```
timfel, line 934: I would write "Wait until force number 1 is completed and ready for orders" to make it clear what exactly we are waiting for

src/ai/script_ai.cpp
SimoneStarace: > It just returns if the unit is needed or it's already existing?
Yes but I think I should put the example inside a function where it's executed by the A.I.
Something like this:
`function() return AiNeed(AiCityCenter()) end`

src/ai/script_ai.cpp
SimoneStarace: Maybe this is better?
`function() return AiSet(AiWorker(), 2) end`

src/ai/script_ai.cpp
SimoneStarace: I was thinking like this:
```
-- Make the force number 1
AiForce(1,{"unit-footman",2, "unit-archer", 1})
-- Wait until force number 1 is completed and ready for orders
AiWaitForce(1)
```
Maybe is cleaner than the previous example.

SimoneStarace:
I have changed some code examples. Maybe those are better than the old ones?

```
/**
** <b>Description</b>
**
**  Need a unit.
**
**  @param l  Lua state.
**
**  @return   Number of return values
**
** Example:
**
** <div class="example">
** 	<code>
**		local ai_function = {
**			-- Say to the A.I that it needs a Town Hall/Great Hall
**    		function() return AiNeed(AiCityCenter()) end,
**		}
** 	</code>
** </div>
*/
static int CclAiNeed(lua_State *l)
```

```
/**
** <b>Description</b>
**
**  Set the number of units.
**
**  @param l  Lua state
**
**  @return   Number of return values
**
** Example:
**
** <div class="example">
** 	<code>
**		local ai_function = {
**			function() return AiSet("unit-peasant",3) end, 
**		}
** 	</code>
** </div>
*/
static int CclAiSet(lua_State *l)
```

```
/**
** <b>Description</b>
**
**  Wait for a force ready.
**
**  @param l  Lua state.
**
** Example:
**
** <div class="example">
** 	<code>
**		local ai_function = {
**			-- Make the force number 1
**			function() return AiForce(1,{"unit-footman",2, "unit-archer", 1}) end,
**			-- Wait until force number 1 is completed and ready for orders
**			function() return AiWaitForce(1) end,
**		}
** 	</code>
** </div>
*/
static int CclAiWaitForce(lua_State *l)
```

PR #345, opened 2021-06-20
Added some code examples.

SimoneStarace:


src/map/script_map.cpp
```diff
@@ -166,9 +177,22 @@ static int CclCenterMap(lua_State *l)
...
 **  @param l  Lua state.
+**
+** Example:
+**
+** <div class="example">
+** 	<code>
+**		  -- Start view for player 0.<br>
+**		  SetStartView(0, 25, 12)<br>
+**		  -- Start view for player 1.<br>
+**		  SetStartView(0, 71, 38)
```
timfel, line 193: I think this should be `SetStartView(1, 71, 38)` :)

src/ui/script_ui.cpp
```diff
@@ -221,9 +273,20 @@ static int CclSetVideoResolution(lua_State *l)
...
+**
 **  Get the video resolution.
 **
 **  @param l  Lua state.
+**
+** Example:
+**
+** <div class="example">
+** 	<code>
+**		resolution = GetVideoResolution()
```
timfel, line 286: here maybe this example would be better:
```
width,height = GetVideoResolution()
print("Resolution  is " .. width .. "x" .. height)
```

src/map/script_map.cpp
SimoneStarace: https://github.com/Wargus/stratagus/commit/72f4bce9a18b577169f05e468e99e81cce150a7d#r52423732

src/ui/script_ui.cpp
SimoneStarace: Fixed the GetVideoResolution example
https://github.com/Wargus/stratagus/commit/72f4bce9a18b577169f05e468e99e81cce150a7d#r52423776

SimoneStarace: Fixed the GetVideoResolution example
https://github.com/Wargus/stratagus/commit/72f4bce9a18b577169f05e468e99e81cce150a7d#r52423776

PR #346, opened 2021-06-26
Added few code examples

SimoneStarace:
This time I have only added few code examples.


SimoneStarace:
In the generate_index.py I have added a variable that stores the css for code exaples. Now in the website it should appear like this
[](https://postimg.cc/6ybswdQ3)

PR #355, opened 2021-07-15
Fix372 win7

psa-jforestier:
Fixing issue https://github.com/Wargus/wargus/issues/372 to make Stratagus work on Windows Seven (including the 32bits version).
Must be strarted with `-g -W` flags to not use OpenGL and have a Window.

Also fixing some crashes because using unintialized string buffer.

### DIFF
--- a/gameheaders/stratagus-game-launcher.h
+++ b/gameheaders/stratagus-game-launcher.h
@@ -579,7 +579,7 @@ int main(int argc, char * argv[]) {
 		// Use extractor from PATH
 		strcpy(extractor_path, EXTRACTOR_TOOL);
 		if (!detectPresence(extractor_path)) {
-			char msg[BUFF_SIZE * 2];
+			char msg[BUFF_SIZE * 2] = {'\0'};;
 			strcpy(msg, EXTRACTOR_NOT_FOUND);
 			strcat(msg, " (expected at ");
 			strcat(msg, extractor_path);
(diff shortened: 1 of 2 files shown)


timfel:
Thanks for those fixes! I added a the bit to automatically force software rendering and windowed mode when we get into the Windows 7 and older branches, so that hopefully Win7 will work again with the next bugfix release.

c-sanchez:
I would like to know how much the engine has improved and how well it works for Starcraft now.
As far as I can see, there is not much information about it.

timfel:
@c-sanchez no one has worked on the Stargus mod since 2017. It's not compatible with the current engine version. One feature in the engine that is missing, but that @ipochto is interested in is the height level for map tiles. Besides that, many things could already be made to work with lua scripting, but it's a lot of work. Since no one has stepped up to do it, I would consider Stargus abandoned for now.

c-sanchez:
Ah, I understand that it doesn't work with the current starcraft engine, especially since it has probably changed a lot with Remastered.
However, it would be great if at some point it would be fully compatible with the latest classic (and now free) version of startcraft, 1.16.1
But I guess I'm for the same reason you say it's in a very early stage of development and now abandoned :(
A shame really, it would be great now that Starcraft is, partially, a free game.

timfel:
@c-sanchez i didn't mean remastered - Stargus is no longer compatible with the current release of the Stratagus engine. It's really abandoned im the state of 2017

dmitry-bikbaev:
> 
> 
> Thanks for those fixes! I added a the bit to automatically force software rendering and windowed mode when we get into the Windows 7 and older branches, so that hopefully Win7 will work again with the next bugfix release.

This 'next bugfix release' you've mentioned isn't 3.1.2 ? I still have the same problem at Win7 x64 with Wargus 3.1.2. I confirm a proposed by @psa-jforestier workaround to run Wargus portable 3.1.2 through 'stratagus.exe -g -W -a -d c:\Wargusdata', is working great.

psa-jforestier:
3.1.2 seems to not contains this patch. Try the 3.1.3 pre-version : 
https://github.com/Wargus/Wargus/releases/master-builds

dmitry-bikbaev:
Tried 3.1.3. Still the same. Product Name: Wargus / Product Version: 3.1.3 / Crash Address: 766CDF15 / Crash Address (Relative): KERNELBASE.dll!ReadFile+0x138 / Exception Code: C0000005 / Exception Description: Access Violation - Failed to access 00000000 memory address (Write Access)

timfel:
3.1.2 includes this patch. It's simply that the software renderer seems not to be automatically applied.

dmitry-bikbaev:
So how to run Wargus.exe with software renderer applied? May be exists such parameter of Wargus.exe?

timfel:
> So how to run Wargus.exe with software renderer applied? May be exists such parameter of Wargus.exe?

wargus.exe accepts the same parameters as stratagus.exe

PR #358, opened 2021-07-27
Add optimization to not draw text when text is empty 

psa-jforestier:
In mission briefing (or possibly anywhere else), there is no need to render empty strings.

### DIFF
--- a/gameheaders/stratagus-game-launcher.h
+++ b/gameheaders/stratagus-game-launcher.h
@@ -237,6 +237,7 @@ int check_version(char* tool_path, char* data_path) {
 	sprintf(buf, "%s -V", tool_path); // tool_path is already quoted
 	HANDLE g_hChildStd_OUT_Rd = NULL;
 	HANDLE g_hChildStd_OUT_Wr = NULL;
+	DWORD nbByteRead;
 	SECURITY_ATTRIBUTES saAttr;
 	saAttr.nLength = sizeof(SECURITY_ATTRIBUTES);
 	saAttr.bInheritHandle = TRUE;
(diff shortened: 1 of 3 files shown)


gameheaders/stratagus-game-launcher.h
```diff
@@ -257,7 +258,7 @@ int check_version(char* tool_path, char* data_path) {
 		return 1;
 	CloseHandle(piProcInfo.hProcess);
 	CloseHandle(piProcInfo.hThread);
-	ReadFile(g_hChildStd_OUT_Rd, toolversion, 20, NULL, NULL);
+	ReadFile(g_hChildStd_OUT_Rd, toolversion, 20, &nbByteRead, NULL);
```
psa-jforestier, line 261: According to https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile#parameters , the `lpNumberOfBytesRead` parameter must not be NULL in this case.

dmitry-bikbaev: Great !
I noted this problem earlier in April and mailed to TimFel but got no changes. I thought the problem was at the Wargus code , but actually at stratagus.
Problem at Stack Overflow: https://stackoverflow.com/questions/46627499/c-readfile-crash-issue/46627945
> in old windows version ReadFile write to lpNumberOfBytesRead if lpOverlapped is 0. begin from 8.1 (or may be 7, I not check now) ReadFile check lpNumberOfBytesRead for 0 value, and not write to it in this case.
@psa-jforestier How (when) to download compiled binaries with this code changes ?

psa-jforestier: You have to wait for TimFel to merge this change on the main branch of Stratagus (I think he is on vacation, so it will probably be merged in a few days/weeks).

PR #365, opened 2021-11-11
enable LUA debugging

andreas-volz:
now only the LUA debug change as PR

### DIFF
--- a/src/stratagus/script.cpp
+++ b/src/stratagus/script.cpp
@@ -2145,9 +2145,11 @@ void InitLua()
 	// For security we don't load all libs
 	static const luaL_Reg lualibs[] = {
 		{"", luaopen_base},
-		//{LUA_LOADLIBNAME, luaopen_package},
 		{LUA_TABLIBNAME, luaopen_table},
-		//{LUA_IOLIBNAME, luaopen_io},
+#ifdef DEBUG
+		{LUA_IOLIBNAME, luaopen_io},
+		{LUA_LOADLIBNAME, luaopen_package},
+#endif
 		{LUA_OSLIBNAME, luaopen_os},
 		{LUA_STRLIBNAME, luaopen_string},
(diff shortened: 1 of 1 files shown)


timfel:
Don't we also need this file somewhere? https://github.com/pkulchenko/MobDebug/blob/master/src/mobdebug.lua

I'm thinking we can just include it and load it eagerly in the debug mode. Include it like we include shaders like this in a `mobdebug.lua` that just lives in the `src/stratagus/` folder
```cpp
R"LUACODE(
-- Copied from pkulchenko/MobDebug src/mobdebug.lua at commit 11d8f83baecac9c39d7868d8fbc74297ca2efbea
--
-- MobDebug -- Lua remote debugger
-- Copyright 2011-20 Paul Kulchenko
-- Based on RemDebug 1.0 Copyright Kepler Project 2005
--
 [... and the entire file ...]
)LUACODE"
```

And then use it in the InitLua function your code before the `tolua_stratagus_open` call like this:
```cpp
#ifdef DEBUG
	static const char* mobdebug =
#include "./modebug.lua"
;
	int status = luaL_loadbuffer(Lua, mobdebug, strlen(modebug), "mobdebug.lua");
	if (!status) {
		status = lua_call(Lua, 0, 1);
		if (!status) {
			lua_setglobal(Lua, "mobdebug")
		}
	}
	report(status, false);
#endif
```

timfel:
This way it's easy for any game to use, by just putting `mobdebug.start()` anywhere

andreas-volz:
Yes, might work as you proposed. Another possibility would be to add it to each apps script folder like this:
https://github.com/andreas-volz/stargus/blob/master/scripts/mobdebug.lua

timfel:
It's merged, mobdebug is included, so you can now always just call `mobdebug.start()` anywhere in any lua script in a debug build.

PR #369, opened 2021-11-24
Vendor in Lua dependencies to fix issues on Fedora 35.

### LINKED ISSUE #366: Consider Vendoring in Older Dependencies.
Currently this project and Wargus depend on some older libraries, namely lua5.1, tolua++(https://github.com/LuaDist/toluapp), and StackTrace(https://sourceforge.net/p/stacktrace/code/HEAD/tree/). tolua++ and lua5.1 currently have some distro specific patching done which potentially is causing some issues with running stratagus on Fedora.  Outside of distros occasional patches there is almost no development activity that I can see happening.  According to the Lua website(https://www.lua.org/versions.html) "The last release was Lua 5.1.5, released on 17 Feb 2012. There will be no further releases of Lua 5.1." Similarly tolua++ had its last commit on Feb 18, 2013, and StackTrace in 2009.

Since lua5.1, tolua++, and StackTrace are highly unlikely to change much in the future, and the patches are likely to be minimal and easily applied as needed, it might make sense to add a directory into the toplevel directory called `third-party` and include the most recent versions of those libraries from ubuntu/debian with all the most recent patches applied to ensure some consistency in the versions used.

Another slightly larger concern is being able to build stratagus in the future as debian(or any other linux distribution) isn't guaranteed to continue packaging these libraries forever. 

If that seems reasonable, I can try working on getting that moving. Lua5.1 and StackTrace use make,  if this proposal is worth pursuing I can work on converting those to use cmake(there should be existing solutions for lua5.1) and integrate all three as ExternalProjects in the CMakeLists.txt file.

If there are any other libraries for which this would seem reasonable, let me know.

abenedic:
This pull request resolves #366 and https://github.com/Wargus/wargus/issues/394. This request adds in a new folder third-party which contains another folder lua-5.1.5 that is the upstream lua-5.1.5 release with a few special cmake files, and tolua++ from the most recent version.  For building tolua++ I added a directory toluapp-simple that is much easier to reason about with cmake and contains only the source files needed to build tolua++.

There is a new build option `option(BUILD_VENDORED_LUA OFF)` which when set to ON builds the vendored lua libs.


abenedic:
Looks like one of the CI builds failed with `Specify which project or solution file to use because the folder contains more than one project or solution file.` And I am not sure why.

abenedic:
https://help.appveyor.com/discussions/problems/7200-specify-which-project-or-solution-file-to-use-because-the-folder-contains-more-than-one-project-or-solution-file,  seems relevant.

abenedic:
I deleted some directories that we don't need, which contatined some files used with Visual Studio, which solved the issue with AppVeyor.

timfel:
Hm, this doesn't work on Ubuntu:
```
[ 82%] Linking CXX executable ../toluapp
/usr/bin/ld: ../liblua51.a(loslib.c.o): in function `os_tmpname':
loslib.c:(.text+0x1c8): warning: the use of `tmpnam' is dangerous, better use `mkstemp'
/usr/bin/ld: ../liblua51.a(loadlib.c.o): in function `ll_unloadlib':
loadlib.c:(.text+0x18): undefined reference to `dlclose'
/usr/bin/ld: ../liblua51.a(loadlib.c.o): in function `ll_load':
loadlib.c:(.text+0x40): undefined reference to `dlopen'
/usr/bin/ld: loadlib.c:(.text+0x50): undefined reference to `dlerror'
/usr/bin/ld: ../liblua51.a(loadlib.c.o): in function `ll_sym':
loadlib.c:(.text+0x93): undefined reference to `dlsym'
/usr/bin/ld: loadlib.c:(.text+0xa3): undefined reference to `dlerror'
collect2: error: ld returned 1 exit status
make[5]: *** [toluapp-simple/CMakeFiles/toluapp.dir/build.make:175: toluapp] Error 1
make[4]: *** [CMakeFiles/Makefile2:173: toluapp-simple/CMakeFiles/toluapp.dir/all] Error 2
make[3]: *** [Makefile:84: all] Error 2
make[2]: *** [CMakeFiles/Lua51B.dir/build.make:112: lua/src/Lua51B-stamp/Lua51B-build] Error 2
make[1]: *** [CMakeFiles/Makefile2:221: CMakeFiles/Lua51B.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

abenedic:
Looks like an issue linking to libdl, I can try spinning up a vm later and fixing that. Fedora and Ubuntu are just different enough to be annoying.

abenedic:
I made a Dockerfile to test this in:
```
FROM ubuntu:20.04
RUN apt-get update && DEBIAN_FRONTEND=noninteractive  apt-get install -y \
        python3\
       python3-pip\
       cmake\
       libsdl2-dev\
       libsdl2-mixer-dev\
       libsdl2-image-dev\
       libtolua-dev\
       libtheora-dev\
       libvorbis-dev\
       libogg-dev\
      libmng-dev\
       libpng-dev\
      zlib1g-dev

COPY ./stratagus/ /stratagus
WORKDIR /stratagus

RUN mkdir build&& cd build && cmake ..  -DBUILD_VENDORED_LUA=ON && make
```

Using that with the latest commit it builds.

abenedic:
The latest commit silences the warning about tempnam, and adds in basic optimizations for release builds of lua.

timfel:
Hm, interesting, it works fine on 20.04 for me too, but not on my other machine which is still on 18.04. But that's fine, we don't need to force the vendored lua on Ubuntu anyway.

PR #372, opened 2021-11-27
Update tinyfiledialog to most recent version.

### LINKED ISSUE #370: Consider updating the version of tinyfiledialogs used.
Currently we are using version 2.3.6 [April 16, 2016], there have been a large number of fixes to the upstream code. It is now on version v3.8.8 [Apr 22, 2021]. If there are no objections I can try seeing if there are any changes needed to replace the current version.

abenedic:
Tiny File dialog introduced some new functions ins the 3.6 release which makes it easier to use UTF-16 and UTF-8 on Windows. Having those functions will aid in the modernization effort in wartool.  This pull request also includes the vendoring of Lua 5.1 and toluapp since I can't build and test without those. I can try rebasing this on master if needed.  This pull request resolves #370.


abenedic:
I also added in a wchar_t version of GetExtractionLogPath, which is needed for dealing with UTF-16 file names.

PR #378, opened 2021-12-20
Fix compiler detection and flags

AMDmi3:
- Use `CMAKE_<LANG>_COMPILER_ID` to detect compiler as CMake docs suggest
- Support clang properly
- Don't set -Werror (see https://amdmi3.ru/posts/packaging-friendliness-compiler-flags-to-avoid/)


timfel:
Thanks!

Issue #386, opened 2022-02-21
handleGenericMessages compiling error on Debian 11 x86_64


sviscapi:
Dear Stratagus team,

I get the following error when trying to compile Stratagus master branch on Debian 11 x86_64 (GCC 10.2):

```
[60%] Building CXX object CMakeFiles/stratagus.dir/src/network/online_service.cpp.o
/home/samuel/src/stratagus/src/network/online_service.cpp:772:5: error: ‘boolean’ does not name a type; did you mean ‘bool’?
  772 |     boolean handleGenericMessages(Context *ctx, uint8_t msg) {
      |     ^~~~~~~
      |     bool
/home/samuel/src/stratagus/src/network/online_service.cpp: In member function ‘virtual void S2C_CHATEVENT::doOneStep(Context*)’:
/home/samuel/src/stratagus/src/network/online_service.cpp:1696:18: error: ‘handleGenericMessages’ was not declared in this scope
 1696 |             if (!handleGenericMessages(ctx, msg)) {
      |                  ^~~~~~~~~~~~~~~~~~~~~
/home/samuel/src/stratagus/src/network/online_service.cpp: In member function ‘virtual void S2C_ENTERCHAT::doOneStep(Context*)’:
/home/samuel/src/stratagus/src/network/online_service.cpp:1722:21: error: ‘handleGenericMessages’ was not declared in this scope
 1722 |                 if (handleGenericMessages(ctx, msg)) {
      |                     ^~~~~~~~~~~~~~~~~~~~~
/home/samuel/src/stratagus/src/network/online_service.cpp: In member function ‘virtual void S2C_CREATEACCOUNT2::doOneStep(Context*)’:
/home/samuel/src/stratagus/src/network/online_service.cpp:1787:21: error: ‘handleGenericMessages’ was not declared in this scope
 1787 |                 if (handleGenericMessages(ctx, msg)) {
      |                     ^~~~~~~~~~~~~~~~~~~~~
/home/samuel/src/stratagus/src/network/online_service.cpp: In member function ‘virtual void S2C_LOGONRESPONSE2::doOneStep(Context*)’:
/home/samuel/src/stratagus/src/network/online_service.cpp:1854:21: error: ‘handleGenericMessages’ was not declared in this scope
 1854 |                 if (handleGenericMessages(ctx, msg)) {
      |                     ^~~~~~~~~~~~~~~~~~~~~
[...]

PR #387, opened 2022-02-26
Changed boolean into bool in online_service.cpp

sviscapi:
It fixes https://github.com/Wargus/stratagus/issues/386 for me


timfel:
Thanks!

PR #393, opened 2022-03-24
Allow build without OpenMP

AMDmi3:
Related to https://github.com/Wargus/wargus/issues/409#issuecomment-1076556344

### DIFF
--- a/src/map/fow.cpp
+++ b/src/map/fow.cpp
@@ -36,7 +36,9 @@
 
 #include <string.h>
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


AMDmi3:
May do. Would it be OK to introduce `src/include/openmp.h` and hide everything there?

timfel:
> May do. Would it be OK to introduce `src/include/openmp.h` and hide everything there?

Fine by me 🙂

timfel:
I've since fixed this in the way that the android port does this.

PR #398, opened 2022-06-05
Making UML Class Diagrams

SimoneStarace:
I started making the UML Class Diagrams of the Stratagus project by using [Umbrello](https://umbrello.kde.org/), which is also the software that can be used to see the diagram in the file.

The aim is to make class diagrams for each folder in the src folder, since the code is structured with folders.

Here is the class diagram only for the [ai](https://github.com/Wargus/stratagus/tree/master/src/ai) folder
stratagus ai class diagram

I would like to know if doing this can be useful or not.


timfel:
This could indeed be useful. The doxygen tool can generate class diagrams, too, but it just includes everything which makes them too big and hard to read.

PR #433, opened 2023-02-25
Silence warnings: Change a logical || to a bitwise |

shinra-electric:
Silences the following compiler warning. Was previously being called 33 times: 
```
warning: use of logical '||' with constant operand [-Wconstant-logical-operand]
        int max = selectMax || INT_MAX;
                            ^  ~~~~~~~
note: use '|' for a bitwise operation
```

Someone with better C++ skills than me should review to make sure functionality remains the same.

### DIFF
--- a/src/include/unit_find.h
+++ b/src/include/unit_find.h
@@ -237,7 +237,7 @@ void SelectFixed(const Vec2i &ltPos, const Vec2i &rbPos, std::vector<CUnit *> &u
 	Assert(Map.Info.IsPointOnMap(rbPos));
 	Assert(units.empty());
 	units.reserve(selectMax << 1);
-	int max = selectMax || INT_MAX;
+	int max = selectMax | INT_MAX;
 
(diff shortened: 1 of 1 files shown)


timfel:
The warning is wrong, the logic here is that if selectMax is 0, we want INT_MAX. I guess you could write `int max = selectMax ? selectMax : INT_MAX;` to make the compiler happy.

src/include/unit_find.h
```diff
@@ -237,7 +237,7 @@ void SelectFixed(const Vec2i &ltPos, const Vec2i &rbPos, std::vector<CUnit *> &u
 	Assert(Map.Info.IsPointOnMap(rbPos));
 	Assert(units.empty());
 	units.reserve(selectMax << 1);
-	int max = selectMax || INT_MAX;
+	int max = selectMax | INT_MAX;
```
timfel, line 240:
```suggestion
	int max = selectMax ? selectMax : INT_MAX;
```

PR #435, opened 2023-03-08
Requires Testing - [macOS] Use the Application Support folder for game data

shinra-electric:
Stratagus currently puts data files in `~/Library/Stratagus`, but it should use `~/Library/Application\ Support/Stratagus`

I have tested this using my previously installed game data. After building with these changes, I moved the Stratagus folder into Application Support and was able to launch Wargus through the command line as usual. 

I have not tested with Wartools to see if it extracts data correctly. That will need to be tested before merging.

### DIFF
--- a/gameheaders/stratagus-game-launcher.h
+++ b/gameheaders/stratagus-game-launcher.h
@@ -223,7 +223,7 @@ static void SetUserDataPath(char* data_path) {
 		if (dataDir) {
 			strcpy(data_path, dataDir);
 #ifdef USE_MAC
-			strcat(data_path, "/Library/Stratagus/");
+			strcat(data_path, "/Library/Application\ Support/Stratagus/");
 #else
 			strcat(data_path, "/.local/share/stratagus/");
 #endif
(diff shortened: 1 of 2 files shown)


timfel:
Code change looks good. Maybe you noticed that for Linux I recently changed it to be more correct and put config and data in different directories (`.local/share/stratagus/` in the game launcher for the data, `.config/stratagus` in `parameters.cpp` for savegames and config files). I don't know if on macOS there's a similar convention to split configuration files and data files.

shinra-electric:
> I don't know if on macOS there's a similar convention to split configuration files and data files.

Well, to be perfectly correct I guess the preferences should be in `~/Library/Preferences`, but I think it's valid to keep the game data and preferences together in Application Support. If you look at comparable apps on [Mac Source Ports](https://macsourceports.com) almost all of them use Application Support. From those games I think only OpenMW and fheroes2 separate out their preferences.

shinra-electric:
I tested this with Wargus and extracting from a GOG exe works fine. It correctly copies the game data to Application Support and can find it there when running. 

I can't test Stargus, because Startools is broken on Mac. (Nothing to do with this PR, I've never been able to get it to work)

timfel:
@shinra-electric thank you for testing!

shinra-electric:
@timfel As an added bonus, this fixes launching by just opening the app bundle. Up until now, I have had to launch from Terminal.

PR #446, opened 2023-08-01
Use `nullptr` instead of `NULL`.

Jarod42:
Some modernization by replacing NULL by `nullptr`.


Jarod42:
Yes. :-)

timfel:
> Yes. :-)

That's nice :) I don't have much time and had to announce I cannot maintain this anymore some months ago. But there's some people that were still very active on Discord (not so much now, since they lost their developer :D) - but Stargus was making really good progress, and we had some fun with the Battle.net support playing War1 games with Lagi's new units and rebalanced stats and I used it to play the War2 coop campaign that SimoneStarace made with my brother.

Jarod42:
I will probably do modernization stuff/clean up things for now.
I no longer have direct access to repo now though :-/

PR #476, opened 2023-08-22
Use `from_chars` 

Jarod42:
(wrapped into `to_number`) for all string to int conversion.


Jarod42:
Our app image is not really C++17. :-/
I have to make my code C++14+ compatible though.

PR #479, opened 2023-08-29
[WIP] Add `DebugPrint` for not found resource by ident.

Jarod42:
So don't forget `-p` command line argument to see typos with identifiers.

### DIFF
--- a/src/unit/upgrade.cpp
+++ b/src/unit/upgrade.cpp
@@ -462,7 +462,6 @@ int UnitTypeIdByIdent(std::string_view ident)
 	if (type) {
 		return type->Slot;
 	}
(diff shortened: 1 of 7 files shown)


src/unit/upgrade.cpp
```diff
@@ -462,7 +462,6 @@ int UnitTypeIdByIdent(std::string_view ident)
 	if (type) {
 		return type->Slot;
 	}
-	DebugPrint(" fix this %s\n" _C_ ident.data());
```
Jarod42, line 465: `UnitTypeByIdent(ident);` already reports issue.

Jarod42:
Too many false positive, creating new resource check existence with those functions...

PR #554, opened 2023-10-29
Try to fix CI due to the update of linuxdeploy:

Jarod42:
Change ICONDIR from /usr/share/icons/default/64x64/ to /usr/share/icons/hicolor/64x64/


Jarod42:
Fixed by https://github.com/Wargus/stratagus/pull/558

PR #581, opened 2023-11-24
Ipochto/highground fixes

ipochto:
- Some fixes to make highgrounds-related wargus's scripts works


src/include/map.h
```diff
@@ -129,6 +129,9 @@ class CMapInfo
 
 	bool IsPointOnMap(const Vec2i &pos) const { return IsPointOnMap(pos.x, pos.y); }
 
+	bool IsHighgroundsEnabled() const { return Highgrounds; }
+	void EnableHighgrounds(const bool enable = true) { Highgrounds = enable; }
```
Jarod42, line 133:
```suggestion
	void EnableHighgrounds(bool enable = true) { Highgrounds = enable; }
```
(non-ref) parameters don't need to be `const`.

src/include/script.h
```diff
@@ -92,12 +92,24 @@ extern int LuaCall(lua_State *L, int narg, int nresults, int base, bool exitOnEr
   lua_error(l); \
  } while (0)
 
-#define LuaCheckArgs(l, args) \
-	do { \
-		if (lua_gettop(l) != args) { \
-			LuaError(l, "incorrect argument"); \
-		} \
-	} while (0)
+static void LuaCheckArgs(lua_State *l, const int args)
```
Jarod42, line 95:
```suggestion
inline void LuaCheckArgs(lua_State *l, const int args)
```
or move definition in cpp. (apply to below function too).

src/map/mapfield.cpp
```diff
@@ -84,7 +84,14 @@ bool CMapField::IsTerrainResourceOnMap() const
 
 void CMapField::setTileIndex(const CTileset &tileset, const tile_index tileIndex, const int value, const uint8_t elevation, const int subtile /* = -1 */)
 {
-	const CTile &tile = tileset.tiles[tileIndex];
+	uint8_t compShift = 0; // [0..F] in case then current tileset slot length is shorter then map's original
```
Jarod42, line 87:
```suggestion
	uint8_t compShift = 0; // [0..F] in case that current tileset slot length is shorter than map's original
```

src/map/script_map.cpp
```diff
@@ -1056,6 +1056,69 @@ static int CclGetIsGameHoster(lua_State *l)
...
+** Example:
+**
+** <div class="example"><code><strong>PresentMap</strong>("Map description", 1, 128, 128, 17)</code></div>
+*/
+static int CclPresentMap(lua_State *l)
+{
+	LuaCheckArgs_min(l, 5);
+
+	Map.Info.Description = LuaToString(l, 1);
+	// Number of players in LuaToNumber(l, 3); // Not used yet.
```
Jarod42, line 1075:
```suggestion
	// Number of players in LuaToNumber(l, 2); // Not used yet.
```

src/map/script_map.cpp
```diff
@@ -1056,6 +1056,69 @@ static int CclGetIsGameHoster(lua_State *l)
...
+/**
+** <b>Description</b>
+**
+** Set basic map caracteristics.
+**
+**  @param l  Lua state.
+**
+** Example:
+**
+** <div class="example"><code><strong>PresentMap</strong>("Map description", 1, 128, 128, 17)</code></div>
```
Jarod42, line 1068:
```suggestion
** <div class="example"><code><strong>PresentMap</strong>("Map description", PlayerCount, Width, Height, uid_number [, "highgrounds-enabled"])</code></div>
```

src/map/script_map.cpp
```diff
@@ -1056,6 +1056,69 @@ static int CclGetIsGameHoster(lua_State *l)
...
+	// Number of players in LuaToNumber(l, 3); // Not used yet.
+	Map.Info.MapWidth = LuaToNumber(l, 3);
+	Map.Info.MapHeight = LuaToNumber(l, 4);
+	Map.Info.MapUID = LuaToNumber(l, 5);
+	
+	if(LuaGetArgsNum(l) >= 6) {
+		const std::string_view highgrounds = LuaToString(l, 6);
+		if (highgrounds == "highgrounds-enabled") {
+			Map.Info.EnableHighgrounds();
+		}
```
Jarod42, line 1084:
```suggestion
		} else {
			LuaError(l, "Unknown value %s\n", highgrounds.data());
		}
```

src/map/script_map.cpp
```diff
@@ -1056,6 +1056,69 @@ static int CclGetIsGameHoster(lua_State *l)
...
+**
+** <div class="example"><code>-- Load map setup from file
+**		<strong>DefineMapSetup</strong>("Setup.sms")</code></div>
+*/
+static int CclDefineMapSetup(lua_State *l)
+{
+	LuaCheckArgs(l, 1);
+	Map.Info.Filename = LuaToString(l, 1);
+
+	return 1;
```
Jarod42, line 1120:
```suggestion
	return 0;
```
Nothing is returned in Lua...

src/include/script.h
ipochto: I was surprised )
https://stackoverflow.com/questions/22102919/static-vs-inline-for-functions-implemented-in-header-files
Thanks.

src/map/script_map.cpp
ipochto: Not sure about this returns.
You mean we should return `1` if we pass something back to lua and `0` otherwise?

src/editor/editloop.cpp
```diff
@@ -1,4 +1,4 @@
-﻿//       _________ __                 __
+//       _________ __                 __
```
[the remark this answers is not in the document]
ipochto: Sorry, this type of changes are accidental.

src/include/map.h
ipochto: Fixed https://github.com/Wargus/stratagus/pull/581/commits/f6231bfe97438217fbac7e2435577fb5904a61a3

ipochto:
@Jarod42, Sorry I'm not so familiar with github's routine, so it turn out that I'm fixed your suggestions twise - via github and locally.

src/guichan/include/guichan/listmodel.h
```diff
@@ -86,6 +86,15 @@ namespace gcn
...
         virtual std::string getElementAt(int i) = 0;
 
+        /**
+         * Gets an index of a given element's value (if present)
+         *
+         * @param element a value of elements to find
+         * @return an index of element if present, -1 otherwise.
+         */
+
+        virtual int getIdxOfElement(std::string_view element) = 0;
```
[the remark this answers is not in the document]
ipochto: I'll check it.
Where I can find last version? Maybe it already has these changes.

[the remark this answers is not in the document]
ipochto: fixed: https://github.com/Wargus/stratagus/pull/581/commits/b9f41e6f5fe7715e3329fc4419892aa0dc8887e0

src/editor/editloop.cpp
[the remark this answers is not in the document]
Jarod42: utf8 BOM

src/map/script_map.cpp
Jarod42: The return value correspond to the number of return value (as tuple from lua stack).

src/editor/editloop.cpp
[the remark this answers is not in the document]
ipochto: fixed

src/map/script_map.cpp
```diff
@@ -1056,6 +1056,71 @@ static int CclGetIsGameHoster(lua_State *l)
...
+	Map.Info.EnableHighgrounds(LuaGetArgsNum(l) >= 1 ? LuaToBoolean(l, 1) : true);
+	
+	return 0;
+}
+
+static int CclIsHighgroundsEnabled(lua_State *l)
+{
+	lua_pushboolean(l, Map.Info.IsHighgroundsEnabled());
+	return 1;
+}
```
Jarod42, line 1103: Might probably go in map.pkg (tolua++) instead.

src/ui/widgets.cpp
```diff
@@ -1718,9 +1719,23 @@ void ImageTextField::drawBorder(gcn::Graphics *graphics)
 }
 
 /*----------------------------------------------------------------------------
---  LuaListModel
+--  StringListModel
 ----------------------------------------------------------------------------*/
 
+int StringListModel::getIdxOfElement(std::string_view element)
```
Jarod42, line 1725:
```suggestion
int StringListModel::getIdxOfElement(std::string_view element) const
```

src/ui/widgets.cpp
```diff
@@ -1735,6 +1750,16 @@ void LuaListModel::setList(lua_State *lua, lua_Object *lo)
 	}
 }
 
+int LuaListModel::getIdxOfElement(std::string_view element)
```
Jarod42, line 1753:
```suggestion
int LuaListModel::getIdxOfElement(std::string_view element) const
```

src/ui/widgets.cpp
```diff
@@ -2610,6 +2635,13 @@ int ImageDropDownWidget::getSelected()
 	return mListBox.getSelected();
 }
 
+std::string ImageDropDownWidget::getSelectedItem()
```
Jarod42, line 2638:
```suggestion
std::string ImageDropDownWidget::getSelectedItem() const
```

src/map/script_map.cpp
ipochto: What's the difference in principle? 

I just did it by analogy, depending on how existing functions with similar functionality were translated into lua. If, for example, what concerns gui was translated through 'tolua++', then almost everything I've encountered concerning engine settings or parameters of the current game was done through `lua_register()` . 

Changing the method is not a problem. I just want to understand for the future on what principle to choose the translation method.

Jarod42: Not sure neither ;-)

to_lua has been added after.

BTW, the flag seems unused in the Engine :-/ ...
Is there incoming PRs next?
How does it work?

ipochto: It used here: https://github.com/Wargus/wargus/pull/446

ipochto: According to this flag we load base or run extended tileset generator to make tiles for highgrounds/ramps/cliffs etc.

Jarod42: Currently
```
      Map.Info.Postamble    = ""
      MapEnableHighgrounds(highgroundsCheckBox:isMarked())
```
With Tolua, it would be more consistent:
```
      Map.Info.Postamble    = ""
      Map.Info.HighgroundsEnabled = highgroundsCheckBox:isMarked()
```
but it seems that Highgrounds is more related to tileset than map.
wonder if it is then the right flag/place.
Shouldn't we choose between summer/summer_highgrounds/swamp/swamp_highgrounds...?

ipochto: >Map.Info.HighgroundsEnabled = highgroundsCheckBox:isMarked()

This is a private member and I don't want to open it in case if we will need to do something else with enabling it in the future. At least the network synchronization may be required. 

I could use Tolua to open `MapEnableHighgrounds()`/`IsHighgroundsEnabled()`, but definitely don't want to make `Map.Info.HighgroundsEnabled` variable as public.

ipochto: > Shouldn't we choose between summer/summer_highgrounds/swamp/swamp_highgrounds...?

Don't breed entities. )

The player simply selects the basic tileset as before, and wargus chooses whether to load the extended version or not. Moreover, if wargus "sees" that the map contains highgrounds, it will simply hide the possibility to select (in dropDownList) the base tileset, for which there is no possibility to generate the extended version (I am talking about swamp specifically - there are certain difficulties with generation).

For the editor, there is a checkbox to enable highgrounds or not.

So the `Map.Info.HighgroundsEnabled` is map's property, on the basis of which the choice of which tileset to load is made

ipochto: I've posted a visual explanation: https://discord.com/channels/780082494447288340/987001693474541570/1179091461820666008

ipochto: Also the map presentation is loaded before tileset. Tileset loads with game start, but map presentation is uploaded to the stage of choosing the game parameters by the player (in the create a new game menu).

PR #586, opened 2023-12-04
Separate highlights of AStar passability from the enable map grid option.

ipochto:
The option is used for debugging purposes and was activated together with enabling the map grid. It was a bit annoying. 
This separates them into two independent options.

Now can be toggled with `SetHighlightPassability(true/false)` from lua (incl. `eval` chit code)

### DIFF
--- a/src/include/viewport.h
+++ b/src/include/viewport.h
@@ -112,11 +112,20 @@ public:
 		return CViewport::ShowGrid;
 	}
 
-	static void EnableGrid(const bool value)
+	static void EnableGrid(bool value)
 	{
 		CViewport::ShowGrid = value;
 	}
 
+	static bool isPassabilityHighlighted()
+	{
+		return CViewport::ShowAStarPassability;
+	}
+
+	static void HighlightPassability(bool value)
+	{
+		CViewport::ShowAStarPassability = value;
+	}
 
 	PixelSize GetPixelSize() const;
 	const PixelPos &GetTopLeftPos() const { return TopLeftPos;}
@@ -155,10 +164,10 @@ public:
 	CUnit *Unit;              /// Bound to this unit
 private:
 	SDL_Surface *FogSurface { nullptr }; /// Texture for fog of war. Viewport sized.
+
 	static bool ShowGrid;
-
+	static bool ShowAStarPassability;
 };
-
 
 //@}
 
--- a/src/map/map_draw.cpp
+++ b/src/map/map_draw.cpp
@@ -45,10 +45,12 @@
 #include "unittype.h"
 #include "ui.h"
 #include "video.h"
+#include "editor.h"
 
 #include <cstdlib>
 
 bool CViewport::ShowGrid = false;
+bool CViewport::ShowAStarPassability = false;
 
 
 CViewport::CViewport() : MapWidth(0), MapHeight(0), Unit(nullptr)
@@ -329,7 +331,7 @@ void CViewport::DrawMapBackgroundInViewport(const fieldHighlightChecker highligh
 			Map.TileGraphic->DrawFrameClip(tile, dx, dy);
 #ifdef DEBUG
 			// AStar passability overlay
-			if (CViewport::isGridEnabled()) {
+			if (CViewport::isPassabilityHighlighted() && Editor.Running == EditorNotRunning) {
 				for (int i = 0; i < graphicTileOffset; i++) {
 					for (int j = 0; j < graphicTileOffset; j++) {
 						if (Map.Fields[sx + j + (mapW * i)].getFlag() & MapFieldUnpassable) {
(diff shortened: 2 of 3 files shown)


src/map/map_draw.cpp
```diff
@@ -329,25 +330,20 @@ void CViewport::DrawMapBackgroundInViewport(const fieldHighlightChecker highligh
 			Map.TileGraphic->DrawFrameClip(tile, dx, dy);
 #ifdef DEBUG
 			// AStar passability overlay
-			if (CViewport::isGridEnabled()) {
-				for (int i = 0; i < graphicTileOffset; i++) {
-					for (int j = 0; j < graphicTileOffset; j++) {
```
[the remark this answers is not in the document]
ipochto: I deleted them by mistake - I didn't recognize right away that Tim implemented highlighting not only for tiles but also for  subtiles (in cases when graphic tile is larger when logical tile). So it will be necessary to return them. For now I've changed the PR status to draft.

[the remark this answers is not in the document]
ipochto: With a little bit of modification, I've reverted highlighting for subtiles. 
This way it will take a bit longer to draw, but I don't think it's critical considering that this function is used only for debugging.  But it is noticeably more readable.

src/map/map_draw.cpp
```diff
@@ -260,21 +260,21 @@ void CViewport::DrawMapGridInViewport() const
 /// Highlight overlays when size of logical and graphical tiles are identical
 void CViewport::HighlightGraphicalTileInViewport(int dx, int dy, uint32_t color /* = ColorRed */, uint8_t transparency /* = 64*/) const
 {
-	Video.FillTransRectangleClip(color, dx, dy, PixelTileSize.x, PixelTileSize.y, transparency);
+    Video.FillTransRectangleClip(color, dx, dy, PixelTileSize.x, PixelTileSize.y, transparency);
```
Jarod42, line 264: Actually, you break indentation style, we use tab as tabulation, (and space for alignment).
I would probably submit .clangformat now that 3rd party code is in third-party directory :-)

src/include/viewport.h
```diff
@@ -126,6 +135,13 @@ class CViewport
 	void Set(const PixelPos &mapPixelPos);
 	/// Draw the map grid for dubug purposes
 	void DrawMapGridInViewport() const;
+	/// Highlight overlays when size of logical and graphical tiles are identical
+	void HighlightGraphicalTileInViewport(int dx, int dy, uint32_t color = ColorRed, uint8_t transparency = 64) const;
+	/// Highlight overlays when size of logical and graphical tiles are differ
```
Jarod42, line 140:
```suggestion
	/// Highlight overlays when size of logical and graphical tiles differ
```
or "are different"

src/map/map_draw.cpp
```diff
@@ -256,6 +258,25 @@ void CViewport::DrawMapGridInViewport() const
...
 	}
 }
+/// Highlight overlays when size of logical and graphical tiles are identical
+void CViewport::HighlightGraphicalTileInViewport(int dx, int dy, uint32_t color /* = ColorRed */, uint8_t transparency /* = 64*/) const
+{
+	Video.FillTransRectangleClip(color, dx, dy, PixelTileSize.x, PixelTileSize.y, transparency);
+}
+
+/// Highlight overlays when size of logical and graphical tiles are differ
+void CViewport::HighlightGraphicalTileInViewport(int dx, int dy, int graphicTileOffset, int sx,
```
Jarod42, line 268:
```suggestion
void CViewport::HighlightGraphicalTileInViewport(int dx, int dy, int graphicTileSize,
```
It seems to be **size** and not **offset**.

I think we can get rid of `sx` (which is for "optimization", as `sx = dx + dy * Map.Info.MapWidth`):
`Map.Fields[sx + j + (Map.Info.MapWidth * i)])` -> `Map.Fields(dx + i, dy + j)`

ipochto: > sx = dx + dy * Map.Info.MapWidth

Not really. `dx` and `dy` are in pixels, `sx` is the index of the top left tile which is at `(dx, dy)`.

ipochto: Anyway, I doubt the while thing about subtiles is working right:

We set a tile from the map file, with `SetTile(graphicTileIdx, tilePos, value, elevation)`
Say we have 4 logical tiles in a row per graphical tile (16 subtiles per whole graphical tile). 
1. And call `SetTile()` for tile with coordinates (0,0).
then here:
https://github.com/Wargus/stratagus/blob/a1bd75ec845eb36b6a55dcc2fe1a174a21f1b395/src/map/script_map.cpp#L805
we will call `*Map.Field(Vec2i(pos.x + j, pos.y + i))` for [0..3, 0..3].
2. Then we want to set next tile `SetTile()` for tile with coordinates (0,1) (call it from a map file). 
And we will call `*Map.Field(Vec2i(pos.x + j, pos.y + i))` for [0..3, 1..4]. So the the `y` range [1..3] will touched twice. And so on.

ipochto: or did I misunderstand the concept

timfel: Yes, I just checked in stargus. `SetTile` x and y are in logical tile coordinates, but since you use it to set the graphical tile, you should skip however many logical tiles are within a graphical tile.

ipochto: > iirc (I may be misremembering) you just shouldn't call `SetTile(0, 0, ...)` and `SetTile(0, 1, ...)` if you have subtiles in e.g. stargus. You should call `SetTile(0, 0, ...)` and then `SetTile(0, 4, ...)` and so on.

Oh, thank you. 
Then there is a need to refine the editor - it doesn't support saving maps with subtiles right now. And check the map conversion code in stargus. 

And in general, I like the idea of subtiles for wargus too.

PR #587, opened 2023-12-06
Add .clang-format to have consistent formatting.

Jarod42:


ipochto:
Maybe increase the column limit?

PR #597, opened 2023-12-26
Fix null pointers check

gromchek:
Added checks to few pointers


src/ai/script_ai.cpp
```diff
@@ -1625,11 +1625,12 @@ static int CclDefineAiPlayer(lua_State *l)
 		if (value == "ai-type") {
 			const std::string_view aiName = LuaToString(l, j + 1);
 			CAiType *ait = GetAiTypesByName(aiName);
-			if (ait == nullptr) {
+			if (ait) {
+				ai.AiType = ait;
+				ai.Script = ait->Script;
+			} else {
 				LuaError(l, "ai-type not found: %s", aiName.data());
```
Jarod42, line 1632: `LuaError` should not `return`, so reordering code is not required.
Same for src/unit/depend.cpp

gromchek: Can you please describe what do you mean by "LuaError should not return"?

Jarod42: As `abort`/`exit`, `LuaError` stops the program.

gromchek: Thanks. I thought about that but decided to clarify.

PR #612, opened 2024-02-25
Change `CColor` to be opaque (alpha=255) instead of transparent

Jarod42:
Should fix issue #596 with font display

PR #616, opened 2024-03-01
Update third-party with guichan patch updated.

Jarod42:
Fix blending with alpha, making wargus "act. I" not black background (see https://github.com/Wargus/stratagus/pull/612#issuecomment-1963831913).


zzam:
"act. I" now has correct background visible.

PR #625, opened 2024-03-05
Fix some load/save issues with Triggers

zzam:
Fix some load/save issues with triggers
Fixes:
* Load will create the original number of triggers (not doubled)
* Load: No bad memory access checking for removed triggers
* Load: removed triggers are now really removed
* Save: The removed-status is correctly saved even if it was from loading a file

### DIFF
--- a/src/game/trigger.cpp
+++ b/src/game/trigger.cpp
@@ -614,8 +614,6 @@ void SaveTriggers(CFile &file)
 	}
 
 	file.printf("\n");
-	file.printf("if (Triggers ~= nil) then assert(loadstring(Triggers))() end\n");
-	file.printf("\n");
 }
 
 /**
--- a/src/game/trigger.cpp
+++ b/src/game/trigger.cpp
@@ -428,7 +428,7 @@ static int CclAddTrigger(lua_State *l)
 	}
 
 	const int i = lua_rawlen(l, -1);
-	if (!ActiveTriggers.empty() && !ActiveTriggers[i / 2]) {
+	if (i / 2 < ActiveTriggers.size() && !ActiveTriggers[i / 2]) {
 		lua_pushnil(l);
 		lua_rawseti(l, -2, i + 1);
 		lua_pushnil(l);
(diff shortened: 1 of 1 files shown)


src/game/trigger.cpp
```diff
@@ -461,10 +474,20 @@ static int CclSetActiveTriggers(lua_State *l)
 {
 	const int args = lua_gettop(l);
 
+	lua_getglobal(Lua, "_triggers_");
```
Jarod42, line 477:
```suggestion
	lua_getglobal(l, "_triggers_");
```

src/game/trigger.cpp
```diff
@@ -461,10 +474,20 @@ static int CclSetActiveTriggers(lua_State *l)
...
 	ActiveTriggers.resize(args);
 	for (int j = 0; j < args; ++j) {
 		ActiveTriggers[j] = LuaToBoolean(l, j + 1);
+		if (j < triggerCount && !ActiveTriggers[j])
+		{
+			TriggerRemoveTrigger(j * 2);
+		}
 	}
+
+	lua_pop(Lua, 1);
```
Jarod42, line 489:
```suggestion
	lua_pop(l, 1);
```
even if `l` is always `Lua` in our code.

src/game/trigger.cpp
```diff
@@ -391,6 +391,19 @@ void ActionStopTimer()
 	GameTimer.Running = false;
 }
 
+/**
+**  Remove a trigger
+**
+**  @param trig  Current trigger
+*/
+static void TriggerRemoveTrigger(int trig)
```
Jarod42, line 399: Not your code, but using `lua_State* l` would avoid global.
Add comment that top stack should be the trigger table.

src/game/trigger.cpp
zzam: fixed

src/game/trigger.cpp
zzam: fixed

src/game/trigger.cpp
zzam: fixed

zzam:
I will do further changes to triggers.
Is it possible to use the LuaCallback from https://github.com/Wargus/stratagus/pull/624 also for triggers?

Jarod42 (approved):

PR #626, opened 2024-03-06
Check each trigger-condition in each cycle

zzam:
Run all check-functions of all triggers in each cycle. Also improve the API of the called functions to get the index of the trigger [0,n-1] instead of an (shifted) index into a lua-table of triggers.

Contrary to the documentation [1] only one trigger condition is run in every Cycle. This state is even saved to a saved game file.

[1] https://github.com/Wargus/stratagus/blob/838f82ec7f31234933c9f999150e17b59c754057/src/game/trigger.cpp#L412-L416

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -533,6 +533,7 @@ set(stratagus_tests_SRCS
 	tests/main.cpp
 	tests/stratagus/test_depend.cpp
 	tests/stratagus/test_luacallback.cpp
+	tests/stratagus/test_trigger.cpp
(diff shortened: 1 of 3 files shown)


Jarod42:
Can you see if unit tests can be added for triggers?
https://github.com/Wargus/stratagus/blob/master/tests/stratagus/test_depend.cpp might be a base.

zzam:
> Can you see if unit tests can be added for triggers? https://github.com/Wargus/stratagus/blob/master/tests/stratagus/test_depend.cpp might be a base.

I will try to write a test.

PR #630, opened 2024-03-12
Make unittests more robust

zzam:
Depend: Fix test error with randomized order (unittype not existing).
Depend: Fix memory leaks.
Trigger: Fix Let tests work with randomized order (GamePaused is not initialized to false).
Trigger: Add more tests of Actions.

### DIFF
--- a/tests/stratagus/test_depend.cpp
+++ b/tests/stratagus/test_depend.cpp
@@ -344,6 +344,7 @@ TEST_CASE("upgrade depends")
 
 TEST_CASE("spell button depend")
 {
+	InitLua_Depend(R"()");
 	const auto requirements = std::string(_("Requirements:\n"));
 	CPlayer player{};
 	ButtonAction button;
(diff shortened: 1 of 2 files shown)


tests/stratagus/test_trigger.cpp
```diff
@@ -31,13 +31,18 @@
...
+#include "interface.h"
+#include "results.h"
 #include "trigger.h"
 #include "script.h"
 
 namespace
 {
 [[nodiscard]] auto InitLuaTrigger(std::string content)
 {
+	GameRunning = true;
```
zzam, line 43: It seems the different parts of stratagus could need better separation.

tests/stratagus/test_depend.cpp
```diff
@@ -357,4 +360,5 @@ TEST_CASE("spell button depend")
 	CHECK(PrintDependencies(player, button) == requirements + "-upgradeMain\n");
 
 	Selected.clear();
+	CleanUnitTypes();
```
Jarod42, line 363: `CleanDependencies();` seems missing, no?

zzam: It did not change anything, but now it looks consistent.

PR #636, opened 2024-03-18
Replace `CConstruction::Frames` from intrusive list to `std::vector<C…

Jarod42:
…ConstructionFrame>`

### DIFF
--- a/src/action/action_built.cpp
+++ b/src/action/action_built.cpp
@@ -77,15 +77,10 @@ void COrder_Built::Save(CFile &file, const CUnit &unit) const /* override */
 	if (this->Finished) {
 		file.printf(" \"finished\", ");
 	}
-	CConstructionFrame *cframe = unit.Type->Construction->Frames;
-	int frame = 0;
-	while (cframe != this->Frame) {
-		cframe = cframe->Next;
-		++frame;
-	}
 	if (this->Worker != nullptr) {
 		file.printf("\"worker\", \"%s\", ", UnitReference(this->Worker).c_str());
 	}
+	const int frame = this->Frame;
 	file.printf("\"progress\", %d, \"frame\", %d", this->ProgressCounter, frame);
 	if (this->IsCancelled) {
 		file.printf(", \"cancel\"");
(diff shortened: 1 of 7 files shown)


src/stratagus/construct.cpp
```diff
@@ -264,15 +258,10 @@ static int CclDefineConstruction(lua_State *l)
...
-				CConstructionFrame **cframe = &construction->Frames;
-				while (*cframe) {
-					cframe = &((*cframe)->Next);
-				}
-				(*cframe) = new CConstructionFrame;
-				(*cframe)->Percent = percent;
-				(*cframe)->File = file;
-				(*cframe)->Frame = frame;
-				(*cframe)->Next = nullptr;
+				auto& cframe = construction->Frames.emplace_back();
```
zzam, line 261: Could the size be set before to avoid re-allocation?

src/action/action_built.cpp
```diff
@@ -331,17 +321,12 @@ void COrder_Built::AiUnitKilled(CUnit &unit)
 }
 
 
-static const CConstructionFrame *FindCFramePercent(const CConstructionFrame &cframe, int percent)
+static std::size_t FindCFramePercent(const std::vector<CConstructionFrame> &cframes, int percent)
 {
-	const CConstructionFrame *prev = &cframe;
+	const auto it =
+		ranges::find_if(cframes, [&](const auto &frame) { return percent < frame.Percent; });
```
zzam, line 327: Is it worth doing binary_search here?
And is it even guaranteed/checked that that Percent-values are ordered correctly?

Jarod42: I tested on wargus and it seems fine...
I will recheck.

src/stratagus/construct.cpp
Jarod42: https://github.com/Wargus/stratagus/pull/637 created.

src/action/action_built.cpp
Jarod42: Well spot, I indeed have bad offset.
I also add pre-require check in https://github.com/Wargus/stratagus/pull/638.

PR #638, opened 2024-03-19
Fix `FindCFramePercent`.

Jarod42:
Add extra checks for construction

### DIFF
--- a/src/action/action_built.cpp
+++ b/src/action/action_built.cpp
@@ -325,8 +325,8 @@ static std::size_t FindCFramePercent(const std::vector<CConstructionFrame> &cfra
 {
 	const auto it =
 		ranges::find_if(cframes, [&](const auto &frame) { return percent < frame.Percent; });
-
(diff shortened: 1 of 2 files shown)


src/action/action_built.cpp
```diff
@@ -325,8 +325,8 @@ static std::size_t FindCFramePercent(const std::vector<CConstructionFrame> &cfra
 {
 	const auto it =
 		ranges::find_if(cframes, [&](const auto &frame) { return percent < frame.Percent; });
```
zzam, line 327: I did not notice before that there is a (backported) ranges-header in here :-)
Nice

zzam, line 327: I played with unit-testing this spot, but it got quite large. Not sure if we should add it.

Jarod42: I didn't look a lot for UTs here :/
Responsability split between Construction and action...

zzam: I will clean the test up and create a PR (maybe today).

PR #639, opened 2024-03-21
Fix glitch of burning building animation

zzam:
Burning animation disappears before building is completely destroyed.
Fix and add unittest.
Simplify storage a bit.
Comment: Percent logic is similar to construction. Code could be combined.

State before:
Burning at 6/400=1.5%:
image

No longer burning at 2/400=0.5%:
image

### DIFF
--- a/src/include/util.h
+++ b/src/include/util.h
@@ -276,10 +276,37 @@ namespace ranges
 		return std::min_element(std::begin(range), std::end(range), cmp);
 	}
 
+	template <typename I, typename S, typename Pred, typename Proj = identity>
+	auto partition_point(I first, S last, Pred pred, Proj &&proj = {})
+	{
+		return std::partition_point(first, last,
+			[&](const auto &elem) {
+				return std::invoke(pred, std::invoke(proj, elem));
+			});
+	}
+
+	template <typename I, typename S, typename Value, typename Comparer = std::less<>, typename Proj = identity>
+	auto lower_bound(I first, S last, const Value &value, Comparer &&comparer = {}, Proj &&proj = {})
+	{
+		return std::lower_bound(first, last, value,
+			[&](const auto &lhs, const auto &rhs) {
+				return std::invoke(comparer, std::invoke(proj, lhs), rhs);
+			});
+	}
+
 	template <typename Range, typename Value>
 	auto lower_bound(Range &range, const Value &value)
 	{
 		return std::lower_bound(std::begin(range), std::end(range), value);
+	}
+
+	template <typename I, typename S, typename Value, typename Comparer = std::less<>, typename Proj = identity>
+	auto upper_bound(I first, S last, const Value &value, Comparer &&comparer = {}, Proj &&proj = {})
+	{
+		return std::upper_bound(first, last, value,
+			[&](const auto &lhs, const auto &rhs) {
+				return std::invoke(comparer, lhs, std::invoke(proj, rhs));
+			});
 	}
 
 	template <typename Range, typename Value, typename Comparer = std::less<>>
--- a/src/include/util.h
+++ b/src/include/util.h
@@ -323,6 +323,12 @@ namespace ranges
 		std::sort(std::begin(range), std::end(range), std::forward<Comparer>(comparer));
 	}
 
+	template <typename Range, typename Comparer = std::less<>>
+	bool is_sorted(Range &range, Comparer &&comparer = {})
+	{
+		return std::is_sorted(std::begin(range), std::end(range), std::forward<Comparer>(comparer));
+	}
+
 }
 
 //@}
--- a/src/include/util.h
+++ b/src/include/util.h
@@ -323,10 +323,13 @@ namespace ranges
 		std::sort(std::begin(range), std::end(range), std::forward<Comparer>(comparer));
 	}
 
-	template <typename Range, typename Comparer = std::less<>>
-	bool is_sorted(Range &range, Comparer &&comparer = {})
+	template <typename Range, typename Comparer = std::less<>, typename Proj = identity>
+	bool is_sorted(const Range &range, Comparer &&comparer = {}, Proj &&proj = {})
 	{
-		return std::is_sorted(std::begin(range), std::end(range), std::forward<Comparer>(comparer));
+		return std::is_sorted(std::begin(range), std::end(range),
+			[&](const auto &lhs, const auto &rhs) {
+				return std::invoke(comparer, std::invoke(proj, lhs), std::invoke(proj, rhs));
+			});
 	}
 
 }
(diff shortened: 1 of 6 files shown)


tests/stratagus/test_missile_fire.cpp
```diff
@@ -0,0 +1,156 @@
...
+//      /   _____//  |_____________ _/  |______     ____  __ __  ______
+//      \_____  \\   __\_  __ \__  \\   __\__  \   / ___\|  |  \/  ___/
+//      /        \|  |  |  | \// __ \|  |  / __ \_/ /_/  >  |  /\___ |
+//     /_______  /|__|  |__|  (____  /__| (____  /\___  /|____//____  >
+//             \/                  \/          \//_____/            \/
+//  ______________________                           ______________________
+//                        T H E   W A R   B E G I N S
+//         Stratagus - A free fantasy real time strategy game engine
+//
+/**@name test_missile_fire.cpp - The test file for missile_fire.cpp / missile.cpp. */
```
ipochto, line 11: Seems to be missed filename change

zzam: fixed

tests/stratagus/test_missile_fire.cpp
```diff
@@ -0,0 +1,172 @@
...
+	SUBCASE("wrong key order") {
+		setFrames({
+			{ 75, M(12)},
+			{ 50, M(11)},
+			{  0, M(10)},
+		});
+
+		CHECK(MissileBurningBuilding(  0) == M(10));
+		CHECK(MissileBurningBuilding( 20) == M(10));
+		CHECK(MissileBurningBuilding( 50) == M(10));
```
zzam, line 166: I am not sure if the test of bad behaviour should be kept.

Jarod42: We should probably have a check in `CclDefineBurningBuilding` that order is correct, and no null value.
And so we can test this checking function instead.

tests/stratagus/test_missile_fire.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+} // anonymous namespace
+
+TEST_CASE("missile_fire")
+{
+	BurningBuildingFrames.clear();
+
+	SUBCASE("size=1") {
+		setFrames({
+			{  0, M(10)},
```
Jarod42, line 60: Maybe have variables for `M(10)`, M(11), ...

tests/stratagus/test_missile_fire.cpp
```diff
@@ -0,0 +1,172 @@
...
+		});
+
+		CHECK(MissileBurningBuilding(  0) == M(10));
+		CHECK(MissileBurningBuilding( 20) == M(10));
+		CHECK(MissileBurningBuilding( 50) == M(10));
+		CHECK(MissileBurningBuilding( 60) == M(10));
+		CHECK(MissileBurningBuilding( 75) == M(10));
+		CHECK(MissileBurningBuilding( 90) == M(10));
+	}
+
```
Jarod42, line 171: Add `BurningBuildingFrames.clear();` to restore global state for other TU.

zzam: fixed

tests/stratagus/test_missile_fire.cpp
zzam: Added and used function is_BurningBuildingFramesSorted.
Null value is allowed to mark not-burning building.

src/include/util.h
```diff
@@ -323,6 +323,12 @@ namespace ranges
 		std::sort(std::begin(range), std::end(range), std::forward<Comparer>(comparer));
 	}
 
+	template <typename Range, typename Comparer = std::less<>>
+	bool is_sorted(Range &range, Comparer &&comparer = {})
```
Jarod42, line 327: `const Range`.

Jarod42: And if you are familiar enough, projection (see above `typename Proj = identity` for `find`)

So we would have:

```
bool is_BurningBuildingFramesSorted()
{
	return range::is_sorted(BurningBuildingFrames, std::greater<>{}, &Frame::Percent);
}
```

src/missile/missile.cpp
```diff
@@ -1297,6 +1297,19 @@ void CleanMissiles()
...
 
+template <typename Range>
+bool is_PercentSortedDecreasing(Range &range)
+{
+	return ranges::is_sorted(
+		range,
+		[](const auto &lhs, const auto &rhs) { return lhs.Percent > rhs.Percent; });
+}
+
+bool is_BurningBuildingFramesSorted()
```
Jarod42, line 1308: I would change Sorted by "OK"/"Valid" and check also that first percent is 0, and last <= 100.

src/include/util.h
zzam: I have seen projections, but is that possible with c++17 ?

Jarod42: I mean like that:
```
template <typename Range, typename Comparer = std::less<>, typename Proj = identity>
bool is_sorted(const Range &range, Comparer &&comparer = {}, Proj &&proj = {}) {
    return std::is_sorted(std::begin(range), std::end(range),
                                    [&](const auto& lhs, const auto& rhs){
                                        return comparer(std::invoke(proj, lhs), std::invoke(proj, rhs);
                                    });
}
```

zzam: That part I got rather quickly after I found the code above.

src/missile/missile.cpp
zzam: done

src/include/util.h
zzam: done. Just needed to use a lambda for the projection (Otherwise the code would not be generic, or ugly)
e.g.
```
typedef typename std::iterator_traits<decltype(range.begin())>::value_type T;
...
then use &T::Percent

```

Jarod42: Mostly because of intermediate `IsPercentDecreasingValid` (used only once).
If code is inlined/moved directly in `IsBurningBuildingFramesValid()`, it would be possible.

Jarod42 (approved):

src/include/util.h
zzam: Sure.
I still wonder if we should combine the percentage-lookup of construction and burning?
That would require the code to be generic/put in header.
How should the signature of such a percentage-lookup function look like?

src/missile/missile.cpp
```diff
@@ -1172,7 +1172,8 @@ int ViewPointDistanceToMissile(const Missile &missile)
 */
 MissileType *MissileBurningBuilding(int percent)
 {
-	for (auto &frame : BurningBuildingFrames) {
+	for (int i = BurningBuildingFrames.size() - 1; i >= 0; --i) {
```
Jarod42: Dangerous to mix signed/unsigned like that.
But changed afterward :-)

PR #647, opened 2024-03-29
Action_built: Add unittest for percentage lookup

zzam:
Testing the percentage lookup code fixed some days ago.

TODO: What else to test in here? and how?

### DIFF
--- a/src/action/action_built.cpp
+++ b/src/action/action_built.cpp
@@ -323,6 +323,10 @@ void COrder_Built::AiUnitKilled(CUnit &unit)
 
 static std::size_t FindCFramePercent(const std::vector<CConstructionFrame> &cframes, int percent)
 {
+	if (cframes.empty())
+	{
+		return -1;
+	}
 	const auto it =
 		ranges::find_if(cframes, [&](const auto &frame) { return percent < frame.Percent; });
 	const std::size_t index = std::distance(cframes.begin(), it);
@@ -340,7 +344,7 @@ void COrder_Built::UpdateConstructionFrame(CUnit &unit)
 	const int percent = this->ProgressCounter / (type.Stats[unit.Player->Index].Costs[TimeCost] * 6);
 	const auto index = FindCFramePercent(type.Construction->Frames, percent);
 
-	if (index != this->Frame) {
+	if (index != -1 && index != this->Frame) {
 		this->Frame = index;
 		const CConstructionFrame &cframe = type.Construction->Frames[index];
 		unit.Frame = (unit.Frame < 0) ? -cframe.Frame - 1 : cframe.Frame;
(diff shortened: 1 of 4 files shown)


src/action/action_built.cpp
```diff
@@ -323,6 +323,10 @@ void COrder_Built::AiUnitKilled(CUnit &unit)
 
 static std::size_t FindCFramePercent(const std::vector<CConstructionFrame> &cframes, int percent)
 {
+	if (cframes.empty())
+	{
+		return -1;
```
Jarod42, line 328: returning `std::optional<std::size_t>` might be an option. that ensures the test for validity.

.editorconfig
```diff
@@ -8,6 +8,7 @@ root = true
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
+trim_trailing_whitespace = true
```
Jarod42, line 11: Better to split PR next time.

tests/stratagus/test_action_built.cpp
```diff
@@ -0,0 +1,251 @@
...
+class tested_COrder_Built : public COrder_Built
+{
+public:
+	using COrder_Built::ProgressCounter;
+
+	int testUpdateConstrFrame_Percent(CUnit &unit, int percent, int initialFrame = 9999)
+	{
+		const int progressOnePercent = unit.Type->Stats[0].Costs[TimeCost] * 6;
+		return testUpdateConstrFrame_Progress(unit, percent * progressOnePercent, initialFrame);
+		}
```
Jarod42, line 46: Not formatted correctly.

tests/stratagus/test_action_built.cpp
```diff
@@ -0,0 +1,251 @@
...
+		Frame = -5; // force re-calculation
+		ProgressCounter = progress;
+		UpdateConstructionFrame(unit);
+		return unit.Frame;
+	}
+};
+
+namespace {
+	void setFrames(std::vector<CConstructionFrame> &cframes,
+				   const std::vector<std::pair<int, int>>& vec)
```
Jarod42, line 60: Would `std::vector<CConstructionFrame> makeFrames(const std::vector<std::pair<int, int>>& vec)` make sense instead?

Jarod42 (approved):

.editorconfig
zzam: Created https://github.com/Wargus/stratagus/pull/648

tests/stratagus/test_action_built.cpp
zzam: fixed

tests/stratagus/test_action_built.cpp
zzam: changed. But does not look too different.

PR #651, opened 2024-04-06
Pathfinding

Jarod42:
Should improve/fix pathfinding with big unit.


src/pathfinder/astar.cpp
```diff
@@ -68,15 +71,20 @@ struct Node {
...
-	int32_t CostFromStart;  /// Real costs to reach this point
-	uint16_t CostToGoal;    /// Estimated cost to goal
-	int8_t InGoal;          /// is this point in the goal
-	int8_t Direction;       /// Direction for trace back
+	// align the matrix, the open set, and the cost to move cache
+	// on 64-byte boundary, in case the memset/memmove operations
+	// of the libc we're using has a 128/256/512bit SIMD vector
+	// instruction branch, since we might be clearing 8M of
+	// memory for a 2048x2048 map
+	alignas(64) int32_t CostFromStart = 0; /// Real costs to reach this point
```
zzam, line 84: Is the goal to align struct Node to 64 bytes?
pahole shows that sizeof(Node) is 64 now (instead of 8).
Isn't the allocator behind vector itself doing a larger alignment for larger allocations?

pahole:
```
struct Node {
private:

        int32_t                    CostFromStart __attribute__((__aligned__(64))); /*     0     4 */
        uint16_t                   CostToGoal;           /*     4     2 */
        int8_t                     InGoal;               /*     6     1 */
        int8_t                     Direction;            /*     7     1 */

        /* size: 64, cachelines: 1, members: 4 */
        /* padding: 56 */
        /* forced alignments: 1 */
} __attribute__((__aligned__(64)));

```

Jarod42: Was indeed to not loose `aligned_malloc(64, AStarMatrixSize)`.
But indeed I did it wrongly.

I think `alignas(std::uint64_t)` was enough, or even no alignment at all.
I will remove it.

PR #656, opened 2024-04-24
Remove `union` in `struct Settings` (type punning is UB in C++).

Jarod42:


Jarod42:
Should be fixed by https://github.com/Wargus/stratagus/pull/660

PR #692, opened 2024-06-09
Display title screen in 4:3

MamiyaOtaru:
displays title screens at the original 4:3 aspect ratio, preserving the art (preventing it from being stretched)

Also exports versions of Video.Width and Height adjusted for 4:3 so games (ie Wargus) can also display menu backgrounds at the original aspect ratio, again preserving the original art

### DIFF
--- a/src/include/video.h
+++ b/src/include/video.h
@@ -328,6 +328,7 @@ public:
 
 	void ClearScreen();
 	bool ResizeScreen(int width, int height);
+	void CalculateAspectRatioCorrectDimensions();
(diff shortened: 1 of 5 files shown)


Jarod42:
Sorry for delay.
You hardcode original ratio. :-/
As alternative, I created #695 which should allow to select the stretching method (as keeping original ratio) (in title screen).
I think the alternative has a better API.

Jarod42:
Superseded by #699
Thanks for code

PR #715, opened 2024-10-14
Migrate (frozen) guichan to guisan

Jarod42:


bjorn:
Note that Guichan isn't exactly frozen, but is now maintained at https://github.com/darkbitsorg/guichan, with recent development happening on https://github.com/darkbitsorg/guichan/tree/0.8 (mostly adding CMake support). I do anticipate an eventual 0.9 release and patches are welcome. :-)

PR #716, opened 2024-10-27
Fix Format function to not cut of last character (fixes game save)

zzam:
This fixes broke game save

The beginning of a broken saved game looks like this:
```
$ zcat ~/.config/stratagus/wc2/save/game.sav.gz|xxd -C|head -5
00000000: 6c6f 6361 6c20 6f6c 6443 7265 6174 6555  local oldCreateU
00000010: 6e69 7420 3d20 4372 6561 7465 556e 6974  nit = CreateUnit
00000020: 006c 6f63 616c 206f 6c64 5365 7452 6573  .local oldSetRes
00000030: 6f75 7263 6573 4865 6c64 203d 2053 6574  ourcesHeld = Set
00000040: 5265 736f 7572 6365 7348 656c 6400 6c6f  ResourcesHeld.lo
```

The newlines are replaced by null-characters so games could not be loaded again.

### DIFF
--- a/tests/stratagus/test_format.cpp
+++ b/tests/stratagus/test_format.cpp
@@ -1,0 +1,48 @@
+//       _________ __                 __
+//      /   _____//  |_____________ _/  |______     ____  __ __  ______
+//      \_____  \\   __\_  __ \__  \\   __\__  \   / ___\|  |  \/  ___/
+//      /        \|  |  |  | \// __ \|  |  / __ \_/ /_/  >  |  /\___ |
+//     /_______  /|__|  |__|  (____  /__| (____  /\___  /|____//____  >
+//             \/                  \/          \//_____/            \/
+//  ______________________                           ______________________
+//                        T H E   W A R   B E G I N S
+//         Stratagus - A free fantasy real time strategy game engine
+//
+/**@name test_format.cpp - Test file for function Format. */
+//
+//      (c) Copyright 2024 by Matthias Schwarzott
+//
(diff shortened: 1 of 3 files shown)


zzam:
I wonder if it is possible to create a larger test that excercises SaveGame+LoadGame to detect such breakage.

tests/stratagus/test_format.cpp
```diff
@@ -0,0 +1,48 @@
...
+	CHECK("12" == Format("12"));
+	CHECK("12\n" == Format("12\n"));
+	CHECK("abc\n" == Format("%s", "abc\n"));
+	CHECK("abc\n" == Format("%s\n", "abc"));
+	CHECK("12345" == Format("%d", 12345));
+	CHECK("         x" == Format("%*s", 10, "x"));
+	std::string result = Format("%*s", 100, "x");
+	CHECK(result[98] == ' ');
+	CHECK(result[99] == 'x');
+	CHECK(result.size() == 100);
```
Jarod42, line 47: Check `size` (with `REQUIRE` ) before accessing 98/99 index.

Jarod42:
> I wonder if it is possible to create a larger test that exercises SaveGame+LoadGame to detect such breakage.

Currently issue is to create such state.

Easier might be using existing savegame, and check that load+save give back same file.
But issue is backward compatibility enforcing/breaking.

tests/stratagus/test_format.cpp
zzam: fixed

PR #736, opened 2025-03-24
Added tool to set a brush tile with picked one from the map.

ipochto:
To set current single tiled type brush with a tile not from tile icons palette, but from any map field under cursor you should:
1. reset selected tile (right click)
2. ctrl+left click on the any tile within map grid.

### DIFF
--- a/src/editor/edmap.cpp
+++ b/src/editor/edmap.cpp
@@ -86,6 +86,34 @@ std::optional<tile_index> CTileIconsSet::getSelectedTile() const
 	return std::nullopt;
 }
 
+bool CTileIconsSet::selectByTile(tile_index tileIdx)
+{
+	const auto prevSelected = selected;
+
+	auto tryToSelect = [this](tile_index idx) {
+		if (const auto it = ranges::find(icons, idx); it != icons.end()) {
+			select(std::distance(icons.begin(), it));
+			return true;
+		}
+		return false;
+	};
+	if (!tryToSelect(tileIdx)) {
+		tryToSelect(Map.Tileset.getFirstOfItsKindTile(tileIdx));
+	}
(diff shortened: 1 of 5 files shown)


src/editor/edmap.cpp
```diff
@@ -86,6 +86,28 @@ std::optional<tile_index> CTileIconsSet::getSelectedTile() const
...
+			select(std::distance(icons.begin(), it));
+			return true;
+		}
+		return false;
+	};
+	if (!tryToSelect(tileIdx)) {
+		tryToSelect(Map.Tileset.getFirstOfItsKindTile(tileIdx));
+	}
+	if (isSelected() && selected != prevSelected) {
+		const uint16_t k = selected / getDisplayedNum();
```
Jarod42, line 104: Is there something which prevent `getDisplayedNum()` to be `0`?

src/map/tileset.cpp
```diff
@@ -780,6 +780,24 @@ std::vector<tile_index> CTileset::queryFirstOfItsKindTiles() const
...
 
+tile_index CTileset::getFirstOfItsKindTile(tile_index tileIndex) const
+{
+	if (getGraphicTileFor(tileIndex) == 0) {
+		ErrorPrint("Requested first tile in the subslot for empty tile: %d\n", tileIndex);
+		return tileIndex;
+	}
+
+	auto result = tileIndex;
+	while ((result % 0xF) != 0) {
```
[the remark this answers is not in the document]
ipochto: You're right - should be `result & 0xF`
fixed

src/editor/edmap.cpp
ipochto: In theory it shouldn't be, but I agree - added handling of such a situation.

Jarod42 (approved):

PR #738, opened 2025-03-26
Restored broken "alt+click" edit tile mode.

ipochto:
Also:
- Fixed permanent brush application while the left button is pressed.
- A little cleanup has been done to make the conditional statements a bit clearer.

### DIFF
--- a/src/editor/editloop.cpp
+++ b/src/editor/editloop.cpp
@@ -93,8 +93,8 @@ static int IconHeight;                      /// Icon height in panels
 static int ButtonPanelWidth;
 static int ButtonPanelHeight;
 
-static bool UnitPlacedThisPress = false;  /// Only allow one unit per press
-static int VisibleUnitIcons = 0;              /// Number of icons that are visible at a time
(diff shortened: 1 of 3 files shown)


src/editor/editloop.cpp
```diff
@@ -93,8 +93,8 @@ static int IconHeight;                      /// Icon height in panels
 static int ButtonPanelWidth;
 static int ButtonPanelHeight;
 
-static bool UnitPlacedThisPress = false;  /// Only allow one unit per press
-static int VisibleUnitIcons = 0;              /// Number of icons that are visible at a time
+static bool ActionDoneThisPress = false;	/// Only allow single action per press
+static int VisibleUnitIcons = 0;			/// Number of icons that are visible at a time
 
 static bool waitForReleaseLeftButton = false;
```
Jarod42, line 99: Is it still used?

Jarod42 (approved):

src/editor/editloop.cpp
ipochto: Ups )
fixed.

PR #743, opened 2025-04-08
Fixed searching for system wide installed lua includes.

ipochto:
Because of these 'include's:
https://github.com/Kitware/CMake/blob/1840c8c61b12a9f46b99018641f514905023133d/Modules/FindLua51.cmake#L39
FindLua51 module searched in:
`include/include/{lua51, lua5.1, lua-5.1, lua}`

Removed additional `include`


Jarod42:
Do you create PR for CMake module directly?

PR #751, opened 2025-07-11
add ptrList vector to unit manager

ghost:
This changes the unit manager so that when the Alloc() function returns a new unit pointer,
it immediately puts the new CUnit* into a vector of pointers to be tracked for deallocation.
During CUnitManager::Init() which runs at the end of every game, the list of released units
is iterated through and each CUnit* is given to delete. The problem is, this doesn't seem to
keep track of all the allocations very well, which is why the game was crashing. I don't really
understand why it happened when mines were depleted, and even then sometimes it wouldn't, but
in any case it's probably a better idea to just put each allocated unit into an array right after
doing new CUnit* to make sure they all get deleted properly. I got a crash once which I can't reproduce
but it's a lot better than crashing at the end of most games.

### DIFF
--- a/src/unit/unit_manager.cpp
+++ b/src/unit/unit_manager.cpp
@@ -63,12 +63,14 @@ void CUnitManager::Init()
 		delete unit;
 	}
 	
+	// Make sure all variables are empty
+	// Initialize the free unit slots
+	// and the released units and pointers
 	//Assert(units.empty());
 	units.clear();
-	// Initialize releasedUnits
-	releasedUnits.clear();	
-	// Initialize the free unit slots
 	unitSlots.clear();
+	releasedUnits.clear();	
+	ptrList.clear();
 }
 
 /**


src/unit/unit_manager.cpp
```diff
@@ -57,15 +57,15 @@ CUnitManager *UnitManager;          /// Unit manager
 void CUnitManager::Init()
 {
 	lastCreated = nullptr;
-	//Assert(units.empty());
-	units.clear();
+
 	// Release memory of units in release list.
-	while (!releasedUnits.empty()) {
-		CUnit *unit = releasedUnits.front();
-		releasedUnits.pop_front();
```
Jarod42, line 65: `releasedUnits` should probably be cleared too

src/include/unit_manager.h
```diff
@@ -74,6 +74,7 @@ class CUnitManager
 	std::vector<CUnit *> units;
 	std::vector<CUnit *> unitSlots;
 	std::list<CUnit *> releasedUnits;
+        std::vector<CUnit *> ptrList;
```
Jarod42, line 77: inconsistent tabulation:
Check your editor configuration (we have clang-format).

PR #764, opened 2026-05-10
Improve automated cross-platform multiplayer integration testing

timfel:
Probably best to review commit by commit.


src/video/graphic.cpp
```diff
@@ -249,6 +249,9 @@ void CGraphic::DrawSubClipCustomMod(int gx, int gy, int w, int h, int x, int y,
 void CGraphic::DrawFrame(unsigned frame, int x, int y,
 						 SDL_Surface *surface /*= TheScreen*/) const
 {
+	if (frame >= frame_map.size()) {
```
Jarod42, line 269: All those checks should probably log on failure, meaning lua scripts is wrong.

src/ui/botpanel.cpp
```diff
@@ -1115,9 +1120,13 @@ void CButtonPanel::DoClicked_SpellCast(int button)
...
 		const bool autocast = ranges::any_of(
-			Selected, [&](const CUnit *unit) { return !unit->AutoCastSpell[spellId]; });
+			Selected, [&](const CUnit *unit) {
+				return static_cast<size_t>(spellId) >= unit->AutoCastSpell.size()
+				       || !unit->AutoCastSpell[spellId];
+			});
 		for (CUnit *unit : Selected) {
-			if (unit->AutoCastSpell[spellId] != autocast) {
+			if (static_cast<size_t>(spellId) >= unit->AutoCastSpell.size()
+			    || unit->AutoCastSpell[spellId] != autocast) {
```
Jarod42: That check seems strange.

src/ai/ai_magic.cpp
```diff
@@ -60,7 +60,8 @@ void AiCheckMagic()
 					return;
 				}
 			}
-			for (unsigned int j = 0; j < SpellTypeTable.size(); ++j) {
+			const size_t spellCount = std::min(SpellTypeTable.size(), unit->Type->CanCastSpell.size());
```
Jarod42, line 65: If sizes mismatch, I would say there is a bug.

src/video/graphic.cpp
timfel: yes, added

src/ai/ai_magic.cpp
timfel: true, I was going through timeless tales and launching a few maps and ran into this.

timfel:
I'm taking a short vacation, I'll revisit this next week

src/ui/botpanel.cpp
timfel: Cleaned up, this was hardening for older savegames, split it so it warns.

src/ai/ai_magic.cpp
timfel: so, indeed this is a bug in timeless tales scripts. i will fix it on that side

timfel:
ok, did some manual cleanup and found a couple more issues. the full tests cannot run on github without the game data though

Jarod42:
Probably better to split and make some dedicated PRs, especially for bug fixes.

timfel:
Should be good to review commit by commit starting from `https://github.com/Wargus/stratagus/pull/764/changes/132553ae13b17f112e775c6bfa63672a1a49ec1b`

PR #766, opened 2026-05-28
Be graceful centering on a group with only buildings in it. Fixes #336

### LINKED ISSUE #336: (Bug Report) War1gus - Screen doesn't move to buildings in a control group
In War1gus, double pressing the control group number of a building doesn't move the screen to it. This issue is not present if the control group is entirely made of units.

timfel:


src/ui/interface.cpp
```diff
@@ -197,19 +197,33 @@ UiCenterOnGroup(unsigned group,
...
+					fallbackPos = unit->GetMapPixelPosCenter();
+				}
 			}
 		}
 	}
 	if (pos.x != -1) {
 		UI.SelectedViewport->Center(pos);
+	} else if (fallbackPos.x != -1) {
+		UI.SelectedViewport->Center(fallbackPos);
 	}
```
Jarod42, line 222: Old formula seems wrongs, it is not the means.

Following untested code seems more correct.
```suggestion
if (units.empty()) {
    return;
}
const auto canBeSelected = [&](const CUnit* unit) { return unit->Type && unit->Type->CanSelect(mode); };
const auto selectableUnitCount = std::ranges::count(units, canBeSelected);
if (selectableUnitCount != 0) {
    const auto pos  = std::accumulate(units.begin(), units.end(), PixelPos{0, 0}, [&](PixelPos& acc, const CUnit* unit) {
        return canBeSelected(unit) ? acc + unit->GetMapPixelPosCenter() : acc;
    }}) / selectableUnitCount;
    UI.SelectedViewport->Center(pos);
} else if (mode != EGroupSelectionMode::SelectAll) {
    const auto pos  = std::accumulate(units.begin(), units.end(), PixelPos{0, 0}, [&](PixelPos& acc, const CUnit* unit) {
        return acc + unit->GetMapPixelPosCenter();
    }}) / units.size();
    UI.SelectedViewport->Center(pos);
}
```

timfel: I updated the code, but using loops, because some units may not be selectable, and we don't use C++20 features, yet, and I didn't want to change that for this PR.

Jarod42 (approved):

PR #768, opened 2026-06-08
Fix multiple issues during network setup.

timfel:
This tests and fixes a few issues during network setup that would break joining or lead to immediate desyncs.

### DIFF
--- a/src/include/netconnect.h
+++ b/src/include/netconnect.h
@@ -113,6 +113,7 @@ extern bool NoRandomPlacementMultiplayer; /// Disable the random placement of pl
 ----------------------------------------------------------------------------*/
 
 extern int FindHostIndexBy(const CHost &host);
+extern bool NetworkIsDedicatedServerHost(const CNetworkHost &host, const CServerSetup &setup);
 extern int NetworkGetPlayerIndexForHost(int hostIndex);
 extern bool NetworkIsLocalSetupInSync(const CServerSetup &state1,
                                       const CServerSetup &state2,
--- a/src/include/netconnect.h
+++ b/src/include/netconnect.h
@@ -34,6 +34,8 @@
 #include "net_message.h"
 #include "network/netsockets.h"
 
+#include <vector>
+
 class CHost;
 
 /*----------------------------------------------------------------------------
@@ -111,6 +113,15 @@ extern bool NoRandomPlacementMultiplayer; /// Disable the random placement of pl
 ----------------------------------------------------------------------------*/
 
 extern int FindHostIndexBy(const CHost &host);
+extern int NetworkGetPlayerIndexForHost(int hostIndex);
+extern bool NetworkIsLocalSetupInSync(const CServerSetup &state1,
+                                      const CServerSetup &state2,
+                                      int hostIndex);
+extern void NetworkApplyClientSetupStateChange(CServerSetup &serverSetup,
+                                               const CServerSetup &clientSetup,
+                                               int hostIndex);
+extern void NetworkCompactHosts();
+extern std::vector<int> NetworkRemoteHostIndices();
 extern void NetworkServerStartGame();       /// Server user has finally hit the start game button
 extern void NetworkGamePrepareGameSettings();
 
--- a/src/network/netconnect.cpp
+++ b/src/network/netconnect.cpp
@@ -1412,6 +1412,7 @@ void CServer::Parse_Map(const int h)
 	switch (networkStates[h].State) {
 		// client has recvd map info waiting for state info
 		case ccs_connected:
+		case ccs_async:
 			networkStates[h].State = ccs_mapinfo;
 			networkStates[h].MsgCnt = 0;
 			[[fallthrough]];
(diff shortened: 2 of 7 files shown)


src/network/netconnect.cpp
```diff
@@ -1675,6 +1713,43 @@ int FindHostIndexBy(const CHost &host)
...
+				if (Hosts[j].IsValid()) {
+					Hosts[i] = Hosts[j];
+					Hosts[j].Clear();
+					anyMoreHosts = true;
+					break;
+				}
+			}
+			if (!anyMoreHosts) {
+				break;
+			}
```
Jarod42: I tend to use `find` for this pattern:
```suggestion
			if (auto it = std::find_if(Hosts + i + 1, Hosts + PlayerMax, &CHost::IsValid); it != Hosts + PlayerMax) {
				Hosts[i] = *it;
				it->Clear();
				break;
			}
```

src/network/netconnect.cpp
```diff
@@ -1842,13 +1951,14 @@ void NetworkServerStartGame()
...
 	const CInitMessage_State statemsg(MessageInit_FromServer, ServerSetupState);
 
-	int hostsToAck = std::max(NetPlayers - 1, 0);
+	const std::vector<int> remoteHostIndices = NetworkRemoteHostIndices();
+	int hostsToAck = remoteHostIndices.size();
 	DebugPrint("Ready, sending InitConfig to %d host(s)\n", hostsToAck);
 	// Send all clients host:ports to all clients.
 	while (hostsToAck) {
 breakout:
 		// Send to all clients, skip ourselves (the server) host in Hosts[0]
```
Jarod42:
```suggestion
		// Send to all clients
```

src/network/network.cpp
```diff
@@ -871,36 +880,39 @@ static void NetworkParseInGameEvent(const unsigned char *buf, int len, const CHo
 */
 void NetworkEvent()
 {
+	static constexpr size_t networkEventBufferSize =
+		CInitMessage_State::Size() > 1024 ? CInitMessage_State::Size() : 1024;
```
Jarod42:
```suggestion
	static constexpr size_t networkEventBufferSize = std::max(CInitMessage_State::Size(), 1024u);
```

src/video/sdl.cpp
```diff
@@ -50,6 +50,7 @@
 #include <climits>
 #include <cmath>
 #include <map>
+#include <sstream>
```
Jarod42: Needed here?

tests/network/test_netconnect.cpp
```diff
@@ -170,12 +196,195 @@ TEST_CASE("CNetworkHost")
...
+	CHECK(static_cast<unsigned>(restored.UserGameSettings) == static_cast<unsigned>(settings.UserGameSettings));
+}
+
+TEST_CASE("Network setup sync tracks ready by host and race by assigned player slot")
+{
+	ClearNetworkHosts();
+	SetHost(3, 7, "client");
+
+	CServerSetup server;
+	CServerSetup local = server;
```
Jarod42: Is copy constructor different than default constructor here (and below)?

src/video/sdl.cpp
timfel: Belongs in other PR, missed that.

tests/network/test_netconnect.cpp
timfel: No different, fixed.

tests/network/test_netconnect.cpp
```diff
@@ -219,6 +237,121 @@ TEST_CASE("Settings bitfield preserves simulation-affecting flags")
 	CHECK(static_cast<unsigned>(restored.UserGameSettings) == static_cast<unsigned>(settings.UserGameSettings));
 }
 
+TEST_CASE("Network setup sync tracks ready by host and race by assigned player slot")
+{
+	ClearNetworkHosts();
+	SetHost(3, 7, "client");
```
Jarod42: I would create some `constexpr` values (hostIndex/PlayerIndex should be sufficient) to avoid magic values, especially as 3 is used as hostIndex and RaceType

src/include/netconnect.h
```diff
@@ -111,6 +113,15 @@ extern bool NoRandomPlacementMultiplayer; /// Disable the random placement of pl
 ----------------------------------------------------------------------------*/
 
 extern int FindHostIndexBy(const CHost &host);
+extern int NetworkGetPlayerIndexForHost(int hostIndex);
```
Jarod42, line 117: As a note, `namespace` might be used instead of prefixing all names.
(Should probably be done in another PR).

src/network/netconnect.cpp
```diff
@@ -1702,6 +1712,37 @@ int FindHostIndexBy(const CHost &host)
...
+					Hosts[j].Clear();
+					anyMoreHosts = true;
+					break;
+				}
+			}
+			if (!anyMoreHosts) {
+				break;
+			}
+		}
+	}
```
Jarod42: [`std::remove_if`](https://en.cppreference.com/cpp/algorithm/remove) might simplify even more that code:
```suggestion
	const auto endValid = std::remove_if(Hosts, Hosts + PlayerMax, [](const auto& host){ return !host.IsValid(); });
	for (auto it != endValid; it != Hosts + PlayerMax; ++it) {
    	it->Clear();
	}
```

Jarod42 (approved):

src/network/netconnect.cpp
timfel: Done

src/include/netconnect.h
timfel: Will do in another PR

tests/network/test_netconnect.cpp
timfel: done

PR #773, opened 2026-06-08
Be nicer to modders with invalid game data and old savegames.

timfel:
Stratagus crashes way too often already, this catches some errors when lua code is slightly wrong or savegames from older engine versions are loaded and at least we do not crash and instead warn.

### DIFF
--- a/src/action/action_still.cpp
+++ b/src/action/action_still.cpp
@@ -226,9 +226,11 @@ static bool MoveRandomly(CUnit &unit)
 */
 bool AutoCast(CUnit &unit)
 {
-	if (!unit.AutoCastSpell.empty() && !unit.Removed) { // Removed units can't cast any spells, from bunker)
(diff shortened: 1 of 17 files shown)


Jarod42:
It seems it requires another PR, can you rebase it so CIs pass

src/action/action_still.cpp
```diff
@@ -52,11 +52,34 @@
...
 #include "video.h"
 
+#include <set>
+
 enum {
 	SUB_STILL_STANDBY = 0,
 	SUB_STILL_ATTACK
 };
 
+static void WarnInvalidAutoCastVectors(const CUnit &unit, const char *context)
```
Jarod42, line 69: The other `WarnInvalidXXX` warn unconditionally, whereas this one does the check itself.
Either adjust naming (`CheckXXX` ?), or change behavior to be consistent.

timfel: Fixed.

Jarod42 (approved):

PR #775, opened 2026-06-08
Various minor improvements to envvar and cmdline handling during startup

timfel:


src/stratagus/stratagus.cpp
```diff
@@ -650,6 +654,34 @@ void ParseCommandLine(int argc, char **argv, Parameters &parameters)
 		break;
 	}
 
+	if (VideoForceFullScreen && argc - optind > 1) {
+		for (int i = optind; i < argc; ++i) {
+			if (parseWindowSize(argv[i])) {
```
Jarod42: Prefix `-W` seems to not be taken into account, is it normal

timfel: Yes :( The problem is really the getopt implementation and us having this non-standard thing with an optional argument, which getopt doesn't support. getopt is allowed to permute non-option arguments and the order is not required to stay the same, so `-W 640x480` may end up having such a size somewhere in the remaining options. So we scan for that, since `-W` is the only weird case like this and the NUMBERxNUMBER format is very unlikely as a map name, this seems the least insane way to deal with this.

Jarod42: Adding comment for that weird stuff might help to understand then.
BTW NUMBERxNUMBER is no so impossible for a map name (which might represent size of the map as bad name).

timfel: I have another idea for a better fix, i'll redo this commit

Jarod42 (approved):

PR #781, opened 2026-06-28
Fix macos bugs with Warcraft II installer extraction

TomAnthony:
This is a bunch of fixes for getting Wargus running on macos, including:

- Find innoextract even when installed by homebrew/ports etc. so the Wargus binary launched as a GUI works
- Fix incomplete file extraction caused by quotes handling on mac
- Fix bug with extraction when not using expansion CD

In case it matters: Claude did the actual code work, with me driving and testing.

Wargus now launches direct from the GUI, where you can select an .exe installation file and it correctly extracts and launches. Tested on macos 26 on ARM.

I haven't got access to Windows, but I think the changes should be safe there but should be tested.


Jarod42 (approved):
Look good to be.
Untested though.

